PR #3, opened 2018-11-29
Add metadata appropriate for a WICG specification.

dbaron:
This should reflect where the spec is being maintained.

I expect this additional metadata will allow working auto-import into
https://github.com/mozilla/standards-positions/

Note that this changes the license information that is generated.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1,6 +1,6 @@
 <pre class="metadata">
 Title: HTML Sanitizer API
-Status: DREAM
+Status: CG-DRAFT
 Group: WICG
 URL: https://wicg.github.io/purification/
 Repository: WICG/purification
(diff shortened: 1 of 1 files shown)


mozfreddyb:
Which change exactly is changing the license information? `Group: WICG`?
I can't seem to find this in the bikeshed docs.

Licensing is not exactly my specialty and I'm not sure what the situation is then versus now.
Would you advise that this needs additional debate?

dbaron:
It is `Group: WICG`, and I don't think it needs additional debate since I think what's it's doing is what's supposed to happen for all WICG specs.

marcoscaceres:
On a related note.... are folks here still pursuing this incubation?

PR #11, opened 2020-06-05
first list of algorithm to sanitizer a document fragment

mozfreddyb:


yoavweiss:
@mozfreddyb - can you join the WICG? The bots are complaining...

mozfreddyb:
I think I need another invitation @yoavweiss

yoavweiss:
I think you just need to click the "join" button on https://www.w3.org/community/wicg/

Issue #30, opened 2020-08-20
toString is reserved identifier.


otherdaniel:
[toString](https://wicg.github.io/sanitizer-api/#sanitizer-api) is a [reserved identifier in WebIDL](https://heycam.github.io/webidl/#idl-names). (See sentence starting with "The identifier of any".) We need to pick a different name.

otherdaniel:
Candidates:
- saneFromString / saneFromFragment
- sanitze / sanitizeToFragment
- stringFrom / fragmentFrom   (Since this is called on a sanitizer instance, the reader already has the context that this is about sanitzing)

shhnjk:
I like sanitze 😊 I always feel bad about assigning string to innerHTML after working on Trusted Types, so something that is different from word _string_ would be nice 😊

shhnjk:
Can people vote if you all prefer _sanitizeToString_ or _sanitizeToHTML_?
If you like _sanitizeToHTML_ please add a 👍 reaction in this comment. If you like _sanitizeToString_, then add a 👎 reaction in this comment.
If you hate both, add a comment 😊

CC: @koto, @mikewest, @cure53

koto:
This depends on https://github.com/WICG/sanitizer-api/issues/20#issuecomment-677797865. If TT would allow creating a `DocumentFragment` for a given sanitizer config, there ideally should be a way to return a `TrustedHTML` instance from said config too. 

I don't think we should even be offering strings as a return value for HTML in new APIs. We know what that led to.

shhnjk:
> This depends on [#20 (comment)](https://github.com/WICG/sanitizer-api/issues/20#issuecomment-677797865). If TT would allow creating a `DocumentFragment` for a given sanitizer config, there ideally should be a way to return a `TrustedHTML` instance from said config too.

Agreed 😊
> I don't think we should even be offering strings as a return value for HTML in new APIs. We know what that led to.

I'm against this idea for Sanitizer API (though this might make sense for other APIs)🙁 The reason being, because we need to be flexible about whatever people decides to add in their _SanitizerConfig_, we do need to return string when we are unsure if the output would cause an XSS. This will ensure that the output is string, and therefore it's still unsafe to assign it as an HTML.

However I have to agree that this would make a confusion that _sanitizeToHTML_ might return a string, which is weird 😂
[...]

Issue #22, opened 2020-07-03
prototype API


mozfreddyb:
For the sake of prototyping, I would like to suggest an API that looks like this (but obviously subject to change):

* You construct a Sanitizer object by supplying options (or accepting the default): `new Sanitizer(options)`
* The sanitizer object has various functions depending on the return type required:
** `sanitize()` returning a DocumentFragment
** `sanitizeToString()` returning a DOMString
* All of those functions should accept documentfragments, documents, strings

WebIDL, scrambled together and likely not entirely correct:

```webidl
typedef (DOMString or DocumentFragment or Document) SanitizerInput;

[Exposed=Window]
interface Sanitizer {
  constructor(optional SanitizerOptions options); // optionality still discussed below
  DocumentFragment sanitize(optional SanitizerInput input);
  DOMString sanitizeToString(optional SanitizerInput input);
};

// unimplemented during prototyping
interface SanitizerOptions {}
```

otherdaniel:
+1

Two rather minor nitpicks:

1, optionality: I'd have done it the other way round.
  - constructor argument: What would that even do? If we're trying to avoid implicit behaviour (see #19), we  can hint in the API that the developer must supply the rules by making it non-optional.
  - sanitize* methods: Making their arguments optional allows the caller to dispense with a null-check (or stringify-able check), so that people can just blindly dump stuff into the sanitizer and be sure to have some resonable output. (Possibly an empty string.)

2, I think web idl mostly uses capitalized names for types (SanitizerOptions, SanitizerInput)

mozfreddyb:
>     * constructor argument: What would that even do? If we're trying to avoid implicit behaviour (see #19), we  can hint in the API that the developer must supply the rules by making it non-optional.

I'm not sure I fully get the ergonomics. Do we expect authors to always spell out the exact set of elements they want to allow? I want us to offer a safe-default that will be (hopefully) adopted by the majority.
I suppose we could set a regularly-re-defined constant `Sanitizer.NO_XSS` that is slightly different per user-agent (e.g., due to element/attribute/event-handler), which would make it both forward-compatble and explicit?
Would that work for you?

> 
>     * sanitize* methods: Making their arguments optional allows the caller to dispense with a null-check (or stringify-able check), so that people can just blindly dump stuff into the sanitizer and be sure to have some resonable output. (Possibly an empty string.)
> 
good point. I'll edit in my initial comment

> 
[...]

PR #31, opened 2020-08-22
Add some integration to Trusted Types

shhnjk:
This PR tries to integrate Sanitizer API with Trusted Types.
I'm not sure this is the right way in terms of WebIDL, but what I'm proposing here is:

1. New _sanitizeToTrustedHTML_ method☺ _sanitizeToTrustedHTML_ will return _TrustedHTML_, and ideally this method will be exposed only if _SanitizerConfig_ is same or stricter than default config.
2. Require _TrustedHTML_ as an argument to _sanitize_ if the _SanitizerConfig_ is weaker than default config.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -100,7 +100,6 @@ let s = new Sanitizer({
   allowAttributes: ['c', 'd', 'e', ...],
   ...
 });
-
 s.sanitize("&lt;img src=x onerror=alert(1)//&gt;"); // returns a 'DocumentFragment'
 s.sanitizeToString("&lt;img src=x onerror=alert(1)//&gt;"); // returns '<img src="x">'
 s.sanitizeToTrustedHTML("&lt;img src=x onerror=alert(1)//&gt;"); // returns 'TrustedHTML'
--- a/index.bs
+++ b/index.bs
@@ -100,9 +100,9 @@ let s = new Sanitizer({
   attributes: ['c', 'd', 'e', ...],
   ...
 });
+s.sanitize("&lt;img src=x onerror=alert(1)//&gt;"); // returns a 'DocumentFragment'
 s.sanitizeToString("&lt;img src=x onerror=alert(1)//&gt;"); // returns '<img src="x">'
 s.sanitizeToTrustedHTML("&lt;img src=x onerror=alert(1)//&gt;"); // returns 'TrustedHTML'
-s.sanitize("&lt;img src=x onerror=alert(1)//&gt;"); // returns a 'DocumentFragment'
 ```
 
 Framework {#framework}
--- a/index.bs
+++ b/index.bs
@@ -121,6 +121,7 @@ Framework {#framework}
   interface Sanitizer {
     constructor(SanitizerConfig? config);
     DocumentFragment sanitize(HTMLString input);
+    DOMString sanitizeToString(DOMString input);
     HTMLString sanitizeToHTML(DOMString input);
 
     // And maybe?
--- a/index.bs
+++ b/index.bs
@@ -100,8 +100,8 @@ let s = new Sanitizer({
   attributes: ['c', 'd', 'e', ...],
   ...
 });
-s.toString("&lt;img src=x onerror=alert(1)//&gt;"); // returns `<img src="x">`
-s.toFragment("&lt;img src=x onerror=alert(1)//&gt;"); // returns a `DocumentFragment`
+s.sanitizeToHTML("&lt;img src=x onerror=alert(1)//&gt;"); // returns '<img src="x">'
+s.sanitize("&lt;img src=x onerror=alert(1)//&gt;"); // returns a 'DocumentFragment'
 ```
 
 Framework {#framework}
@@ -120,8 +120,8 @@ Framework {#framework}
   [Exposed=(Window)]
   interface Sanitizer {
     constructor(SanitizerConfig? config);
-    DOMString toString(DOMString input);
-    DocumentFragment toFragment(DOMString input);
+    [CEReactions] DocumentFragment sanitize(HTMLString input);
+    HTMLString sanitizeToHTML(DOMString input);
 
     // And maybe?
     static DOMString sanitizeToString(DOMString input, optional SanitizerConfig config);
--- a/index.bs
+++ b/index.bs
@@ -120,7 +120,7 @@ Framework {#framework}
   [Exposed=(Window)]
   interface Sanitizer {
     constructor(SanitizerConfig? config);
-    [CEReactions] DocumentFragment sanitize(HTMLString input);
+    DocumentFragment sanitize(HTMLString input);
     HTMLString sanitizeToHTML(DOMString input);
 
     // And maybe?
--- a/index.bs
+++ b/index.bs
@@ -116,13 +116,15 @@ Framework {#framework}
     // ...
     // More things from https://github.com/cure53/DOMPurify/blob/master/src/purify.js#L224
   };
+  typedef [StringContext=TrustedHTML] DOMString HTMLString;
+  typedef (DOMString or HTMLString or DocumentFragment or Document) SanitizerInput;
 
   [Exposed=(Window)]
   interface Sanitizer {
     constructor(SanitizerConfig? config);
-    DocumentFragment sanitize(HTMLString input);
-    DOMString sanitizeToString(DOMString input);
-    HTMLString sanitizeToHTML(DOMString input);
+    DocumentFragment sanitize(optional SanitizerInput input);
+    DOMString sanitizeToString(optional SanitizerInput input);
+    HTMLString sanitizeToHTML(optional SanitizerInput input);
 
     // And maybe?
     static DOMString sanitizeToString(DOMString input, optional SanitizerConfig config);
--- a/index.bs
+++ b/index.bs
@@ -100,7 +100,8 @@ let s = new Sanitizer({
   attributes: ['c', 'd', 'e', ...],
   ...
 });
-s.sanitizeToHTML("&lt;img src=x onerror=alert(1)//&gt;"); // returns '<img src="x">'
+s.sanitizeToString("&lt;img src=x onerror=alert(1)//&gt;"); // returns '<img src="x">'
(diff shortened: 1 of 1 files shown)


mozfreddyb:
I'm not sure I fully understand all of those and the renaming makes it a bit more confusing :confused: 
Can you elaborate further? The Custom Element reactions `[CEReactions]` that you sneaked in come without an explanation as well :-)

mozfreddyb:
I think this pull request needs to be a bit more elaborate.
We'll need
- sanitizeToString() for sites that don't use (and browsers that don't implement) Trusted Types
- sanitizeToHTML() for those that do
- sanitize() for DocumentFragments, as previously

Please also explain why you added `[CEReactions]`, but only for some.

shhnjk:
>The Custom Element reactions [CEReactions] that you sneaked in come without an explanation as well :-)

Explanation of this is already added as following. [CEReactions] indicate that this might throw a Trusted Types violation. See [this](https://w3c.github.io/webappsec-trusted-types/dist/spec/#extensions-to-the-document-interface) for example.
`2. Require TrustedHTML assignment to sanitize if the SanitizerConfig is weaker than default config.`

> sanitizeToString() for sites that don't use (and browsers that don't implement) Trusted Types

I don't think we need this. _HTMLString_ will require _TrustedHTML_ only if Trusted Types is enforced on current document (which includes the case where a browser doesn't support TT).

Let me know if anything else needs explanation 😊

shhnjk:
> Please also explain why you added [CEReactions], but only for some.

I forgot to explain this 😅
_sanitizeToHTML_ shouldn't trigger a TT violation, because the output will be a _TrustedHTML_ or a string (in the case where _SanitizerConfig_ is weaker than default config). If the output was a string, then assigning that string to a dangerous sink would cause a TT violation.

However, we should cause a TT violation in _sanitize_, because appending _DocumentFragment_ to the DOM will never cause a TT violation. So if we don't cause a violation where _SanitizerConfig_ is weaker than default config, and that sanitizer called  _sanitize_ with untrusted data, this will introduce a Trusted Types bypass.

koto:
`[CEReactions]` are not related to Trusted Types logic, it's for [custom elements](https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions). 

You might need them though when creating the nodes in the fragment (e.g. see how element creation calls into custom elements logic - https://dom.spec.whatwg.org/#document), as pretty much any DOM operation needs them, but that's for different reasons.

shhnjk:
LOL, I totally get it wrong then :D

shhnjk:
I'm now convinced that we might want to take @mozfreddyb's https://github.com/WICG/sanitizer-api/pull/31#issuecomment-680715193 with slight modification proposed in https://github.com/WICG/sanitizer-api/issues/30#issuecomment-681658010.

index.bs
```diff
@@ -120,8 +120,9 @@ Framework {#framework}
   [Exposed=(Window)]
   interface Sanitizer {
     constructor(SanitizerConfig? config);
-    DOMString toString(DOMString input);
-    DocumentFragment toFragment(DOMString input);
+    DocumentFragment sanitize(HTMLString input);
```
shhnjk, line 123: Is following a better way to represent that _sanitize_ can take a string (if _SanitizerConfig_ is default config or stricter) but may need a _TrustedHTML_ (if _SanitizerConfig_ is weak AND TT is enforced)?
```suggestion
    DocumentFragment sanitize(HTMLString or DOMString input);
```

otherdaniel:
Generally, looks good!

On the Chromium side, we model Trusted Types (mostly) by having a (non-standard) custom attribute, `[StringContext=TrustedHTML]`, and then we can generate the appropriate checks automatically. We only explicitly use the trusted type when the automatic check generation won't work for some reason.

I like the extended attribute, because this way the IDLs remain closer to standard, and an implementation that doesn't want Trusted Types can just ignore the attribute and mostly everything works. So if this was Blink, I'd write it as something like:
```
DocumentFragment sanitize([StringContext=TrustedHTML] DOMString input);
TrustedHTML sanitizeToHTML(DOMString input);
DOMString sanitizeToString(DOMString input);
```

That said, I'm not sure if we can use it here, since I don't think `[StringContext]` is standard. Can we just reference the TT spec draft?


I suppose we also will need - probably somewhere on the Trusted Types side -  a rule that `trusted-types: 'sanitizer'` enables Sanitizer API in TT.


Reference: [[StringContext] IDL doc](https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/bindings/IDLExtendedAttributes.md#stringcontext_trustedhtml_trustedscript_trustedscripturl_t)
Reference: [[StringContext] use in Blink](https://source.chromium.org/search?q=file:%5C.idl$%20StringContext%3D)

mozfreddyb:
Please build this PR on top of https://github.com/WICG/sanitizer-api/issues/22#issue-650443184, specifically I'd want input like `typedef (DOMString or DocumentFragment or Document) SanitizerInput;`

otherdaniel:
I see. So we want something like:

```webidl
dictionary SanitizerOptions { [... TDB ...] }
typedef (DOMString or DocumentFragment or Document) SanitizerInput;

[Exposed=Window]
interface Sanitizer {
  constructor(optional SanitizerOptions options);

  DocumentFragment sanitize(optional SanitizerInput input);
  DOMString sanitizeToString(optional SanitizerInput input);
}
```

I'm a little confused about how to model TT. If we take the opinion that sanitizers are always safe (and, I presume, some on/off switch in the trusted-types CSP header), then we can just add a method for a new output type:

```webidl
  TrustedHTML sanitizeToHTML(optional SanitizerInput input);
```

If instead we think we may need to do a TT check for (some) methods, we can do something like:

```webidl
  typedef (([StringContext=TrustedHTML] DOMString) or DocumentFragment or Document) UntrustedSanitizerInput;

  DocumentFragment sanitize(optional UntrustedSanitizerInput input);
  TrustedHTML sanitizeToHTML(optional UntrustedSanitizerInput input);
```

In terms of spec mechanics, we do have references for the two new concepts here (the TrustedHTML type and the StringContext attribute):
* StringContext: https://w3c.github.io/webappsec-trusted-types/dist/spec/#webidl-string-context-xattr
* TrustedHTML: https://w3c.github.io/webappsec-trusted-types/dist/spec/#trusted-html

index.bs
shhnjk: Will take _SanitizerInput_ suggested :)

shhnjk:
I've added a _typedef_ for _HTMLString_ so that it'll be cleaner. And then added proposal from https://github.com/WICG/sanitizer-api/issues/22#issue-650443184.
TT check may be required for _sanitize_ , which can be defined by logic later. And ideally, _sanitizeToHTML_ won't be exposed if  _SanitizerConfig_ is weaker than default config (thus browsers that doesn't support won't need to implement  _sanitizeToHTML_).

index.bs
```diff
@@ -116,12 +116,15 @@ Framework {#framework}
     // ...
     // More things from https://github.com/cure53/DOMPurify/blob/master/src/purify.js#L224
   };
+  typedef [StringContext=TrustedHTML] DOMString HTMLString;
+  typedef (DOMString or HTMLString or DocumentFragment or Document) SanitizerInput;
```
otherdaniel, line 120: `DOMString or HTMLString or ...` I suspect this isn't going to work. The extended attribute doesn't really create a new type, and the result of this is that we now have a union with the same type twice, and the bindings won't know how to convert a given value into which of the two `DOMString`s.

(To be honest, I'm not super familiar with WebIDL and the bindings compiler, so maybe my understanding here is flawed...)

So: From the WebIDL (and bindings') perspective, we have three things:
- ` TrustedHTML` (this is the "trusted" type)
- `DOMString` (just a regular string)
- `[ScriptContext=String] DOMString`  (still a regular string)

That last thing is still a regular DOMString, but we're telling the bindings that this value will be used for something HTML-ish. Trusted Types then comes in and generates an additional check & conversion logic when it sees these. But this doesn't really change the underlying type; it's more a hint to the bindings compiler to add additional behaviour to the bindings.

So: If we want all methods to have the same input type(s), then we should probably have a typedef over `DOMString or TrustedHTML or ..` In that case, the bindings won't add special behaviour (and couldn't, because the behaviour would slightly differ depending on the method), and we'd add the checks manually. (No big deal, though.)

Or, we want to rely on the bindings to generate the checks. In that case, we can't quite preserve the symmetry between the methods and need to tell the bindings (via the extended attribute) that for two of these methods the input string will be used in HTML-ish context, while for the string-to-string method it won't.

index.bs
```diff
@@ -116,12 +116,15 @@ Framework {#framework}
...
 
   [Exposed=(Window)]
   interface Sanitizer {
     constructor(SanitizerConfig? config);
-    DOMString toString(DOMString input);
-    DocumentFragment toFragment(DOMString input);
+    DocumentFragment sanitize(optional SanitizerInput input);
+    DOMString sanitizeToString(optional SanitizerInput input);
+    HTMLString sanitizeToHTML(optional SanitizerInput input);
```
otherdaniel, line 127: HTMLString -> TrustedHTML

(This specifies a DOMString return value, with an additional Trusted Type check, which won't really do anything for a return value.)

mozfreddyb: Nit:
I think `HTMLString sanitizeToHTML` will confuse a lot of folks into expecting a string.
Especially as DOMString is very jargon-y for developers. :confused: 
Could we go for `sanitizeToTrustedHTML`, instead?

index.bs
shhnjk: Removed _DOMString_. Hopefully this now makes sense 😊

index.bs
shhnjk: Done 😊

shhnjk:
@mozfreddyb and @otherdaniel, I've addressed all comments. Is this good enough to merge?

shhnjk:
This PR depends on https://github.com/WICG/sanitizer-api/issues/37, as if we aren't exposing `sanitizeToString` then we probably don't want to expose `sanitizeToTrustedHTML` either.

shhnjk:
Looks like #49 will be taking care of this, so closing.

Issue #33, opened 2020-09-10
Semantics of allow vs block vs defaults


otherdaniel:
@iVanlish has thankfully been looking at implementing this, and we noticed that we've never bothered to define (or agree on) how allow-lists & block-lists & the default config would play together. My current idea would be something like so:

1. For each element in the source fragment, we determine an action `keep` | `block` | `drop`.
2. if `keep`, we keep the element.
3. if `block`, we remove the element and replace it with its children, which we'll process in turn
4. if `drop`, we remove the element and all its children from the tree.

Decision making: If the current element is:

1. in the built-in default drop list: `drop`.
2. in the built-in default block list: `block`.
3. in the config's drop list: `drop`.
4. in the config's block list: `block`.
5. not in the config's allow list and the allow list is non-empty: `block`
6. otherwise: `keep`

---

This is a bit different (& more explicit) than the current proposals:
- Instead of a duality allow/block, this now has three behaviours: allow/block/drop.
- In this definition, drop overrides block, and block overrides allow.
- In this definition, the built-in lists override the config. There is no way to override the built-ins.

I think this definition could be analogously extended for atributes, or namespaced elements.

There are several related issues (#29 #17 #18 #26), but none seem to quite answer this question.

---

Generally, I don't feel super strongly. This is more a request to settle on some specific behaviour than a proposal for a specific one. I'd be particularly interested in feedback from practicioners on whether this matches their needs. @mozfreddyb @shhnjk @koto.

iVanlIsh:
I have a slightly different option of the content 
> * In this definition, the built-in lists override the config. There is no way to override the built-ins.

IMHO, built-ins suppose to be an easy way for people who do not want to bother making costumed configs themselves to start with other than to force people to follow our options of what is sensitive.

So I would suppose:
* Drop overrides block & build-in block & allow;
* Block overrides allow;
* Allow overrides build-in block & build-in drop.

otherdaniel:
_(Feedback delivered to me in private:)_ General skepticism about the allow/block/drop. Allow/block might be just fine for the vast majority of use cases.

iVanlIsh:
Sum up a little bit here:

1. There are three options now for the strategy besides a simple allow or drop choice:
* blocks: may have risks for changing tree structure
* neuter: way to keep the text children
* callbacks: significantly complicate for implementation

2. Allow/drop ( other than allow/block ) might be more practical and would have the vast majority in real life cases.
Reasons down here:
[...]

PR #39, opened 2020-10-09
Formalize drop lists (& friends)

otherdaniel:
Specify behaviour of drop, block, and allow lists. This means to formalize the discussion at #33.

Features are:

1, Built-ins are handled first, and thus cannot be overriden.
2, Except dropping an element with all children is handled before built-in element blocking, which should allow us to emulate DOMPurify's KEEP_CONTENT.
(Assuming that the built-ins e.g. block <script> (but not its content), a user could then add <script> into the drop list, meaning that <script> and all children would be dropped.)
3. I've tried to translate this into spec-like algorithms language. First time, so... please look closely.

This would also obsolete #15.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,15 +64,22 @@ Note: The proposal is being developed [here](https://wicg.github.io/sanitizer-ap
 We want to develop an API that learns from the
 [DOMPurify](https://github.com/cure53/DOMPurify) library. In particular:
 
-* The core API should be a single String-to-String method, plus a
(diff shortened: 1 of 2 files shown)


iVanlIsh:
LGTM

iVanlIsh (approved):
LGTM

mozfreddyb (approved):
super embarrassed about the delay. looks good!

PR #40, opened 2020-10-12
Cleanup: Remove un-specced features.

otherdaniel:
This removes a that are potentially useful, but so far were neither speced out nor discussed properly. These should be trivial to add back if we want to support them after all.

Additionally, removing the static methods fixes a bikeshed error about duplicate defintiions of "sanitizerToString".


iVanlIsh (approved):
LGTM

Issue #37, opened 2020-10-06
Rethink how we make sanitizeToString an mXSS-safe method?


shhnjk:
Interesting blog post came out today about recent DOMPurify bypass:
https://research.securitum.com/mutation-xss-via-mathml-mutation-dompurify-2-0-17-bypass/

The blog post ends with interesting conclusion:
>The bypass also made me realize that the pattern of
`div.innerHTML = DOMPurify.sanitize(html)`
is prone to mutation XSS-es by design and it’s just a matter of time to find another instances. I strongly suggest that you pass RETURN_DOM or RETURN_DOM_FRAGMENT options to DOMPurify, so that the serialize-parse roundtrip is not executed.

This is equivalent to our `sanitizeToString`. Maybe we should drop the support of `sanitizeToString` all together?
I think `sanitize` should work for most of cases, and developers can create `sanitizeToString` behavior by themselves if they really need to, but we can choose to not recommend this for mXSS purpose.

WDYT?

koto:
To me, it's rather a yet another example of why sanitizers (the ones that aim to prevent XSS vectors) should not output elements from non-html namespaces.

shhnjk:
That makes sense, but what if destination element is actually not an HTML namespace?
e.g.:
`svg.innerHTML = sanitizer.sanitizeToString(untrusted);`

Maybe we want an optional argument in `creationOptions`, where developers can tell which namespace they plan to assign (default to HTML) and we change parser based on that information?

securityMB:
Since @shhnjk pointed me to this thread on Twitter, here's my two cents:

I think that sanitizing to string in JS is risky from both conceptual and practical level. 

Conceptual, because we lose some of advantages brought by sanitizing in the browser; even the spec admits that serialization and reparsing of the DOM tree is not guaranteed to return the original DOM tree. Hence there's always a risk that a DOM tree that the sanitizer believes to be safe will become dangerous after serialize-reparse. To truly benefit from sanitizing in the browser, we should work on document fragments (or just nodes) so that the sanitizer knows exactly what elements, attributes and text content are in there.

Looking from a practical standpoint, AFAIK most of recent bypasses of sanitizers are result of serialization quirks. For instance, Gareth Hayes found [DOMPurify bypass in Edge](http://www.thespanner.co.uk/2018/07/29/bypassing-dompurify-with-mxss/) because it mutated the `title` tag from:

```
[...]

PR #41, opened 2020-10-12
Remove string examples

otherdaniel:
Follow-up from #37: If we think of the DocumentFragement-based API as the primary sanitizer API, maybe we shouldn't lead our motivating examples with the string version. This strips down the examples a bit, and re-writes them to use fragments instead of strings.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -95,13 +95,21 @@ Examples {#examples}
 --------------------
 
 ```js
+let userControlledInput = "&lt;img src=x onerror=alert(1)//&gt;";
+
+// Create a DocumentFragment from unsanitized input:
+let s = new Sanitizer();
+let sanitizedFragment = s.sanitize(userControlledInput);
+
+// Replace an element's content from unsanitized input:
+element.innerText = "";
+element.appendChild(s.sanitize(userControlledInput));
+
+// Use the sanitizer to handle additional input that might be undesired for
+// a specific application, for example hyperlinks or plugins.
 let s = new Sanitizer({
-  allowElements: ['a', 'b', ...],
-  allowAttributes: ['c', 'd', 'e', ...],
-  ...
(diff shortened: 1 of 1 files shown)


iVanlIsh (approved):
LGTM

index.bs
```diff
@@ -95,13 +95,21 @@ Examples {#examples}
...
 ```js
+let userControlledInput = "&lt;img src=x onerror=alert(1)//&gt;";
+
+// Create a DocumentFragment from unsanitized input:
+let s = new Sanitizer();
+let sanitizedFragment = s.sanitize(userControlledInput);
+
+// Replace an element's content from unsanitized input:
+element.innerText = "";
+element.appendChild(s.sanitize(userControlledInput));
```
mozfreddyb, line 106: Nit: Instead of 
```
element.innerText = "";
element.appendChild(s.sanitize(userControlledInput));
```
You could just do `element.replaceChildren(s.sanitize(userControlledInput));`

index.bs
```diff
@@ -95,13 +95,21 @@ Examples {#examples}
...
+
+// Use the sanitizer to handle additional input that might be undesired for
+// a specific application, for example hyperlinks or plugins.
 let s = new Sanitizer({
-  allowElements: ['a', 'b', ...],
-  allowAttributes: ['c', 'd', 'e', ...],
-  ...
+  blockElements: ['a', 'object', ...],
```
mozfreddyb, line 111: Not sure how we agree on how options are currently passed. Maybe leave this one out and use `new Sanitizer(...)` for now?

mozfreddyb (changes requested):
Looks good! Just two questions :)

index.bs

index.bs

mozfreddyb (approved):
:+1:

PR #47, opened 2020-11-23
Define all references; add examples.

otherdaniel:
This change makes the spec compile without warnings with the bikeshed tool.

This defines all references (in a few places as placeholders), but otherwise doesn't mean to change the semantics. It also adds some exmaples.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -110,15 +110,13 @@ Framework {#framework}
 
 ## Sanitizer API {#sanitizer-api}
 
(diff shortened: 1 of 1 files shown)


mozfreddyb:
thanks :+1: , let's take this and iterate on details in follow-ups if needed

mozfreddyb (approved):

iVanlIsh (approved):
LGTM. Thank you!

PR #49, opened 2020-11-26
Draft integration between Sanitizer API and Trusted Types

otherdaniel:


index.bs
```diff
@@ -176,3 +176,86 @@ Acknowledgements {#ack}
...
+<pre>
+  // In interface Sanitizer:
+  <ins>TrustedHTML sanitizeToTrustedHTML(TrustedTypesSanitizerInput input);</ins>
+</pre>
+
+Since this is - naturally - a new sink, it requires a Trusted Types check.
+But it clearly makes the sense to use this as a way to create Trusted Types
+from untrusted input. To facilitate this, the
+<a href="https://w3c.github.io/webappsec-trusted-types/dist/spec/#trusted-types-csp-directive">`trusted-types` CSP directive</a>
+gets a new keyword `'sanitizer-api'`. If set, the `sanitizeToTrustedHTML` method
```
koto, line 223: You might want to move this to require-trusted-types-for keyword, e.g.

`require-trusted-types-for 'script' 'allow-sanitizer-api'`

This is the directive that controls sinks. `trusted-types` controls policy creation.

index.bs
```diff
@@ -176,3 +176,86 @@ Acknowledgements {#ack}
...
+will accept any DOMString input. That is, it will skip the Trusted Types check
```
koto, line 224: This will need to be formalized, but LGTM.

koto (approved):

index.bs
otherdaniel: Hm. I had seen this as effectively creating implicit policies, hence `trusted-types` rather than `require-trusted-types-for`. I can see how it could also be a seperate class for sinks.

I do think seeing this as a sink class is a little awkward, since having a sanitizer create `TrustedHTML` instances falls into the domain of `script`.

(I don't feel super strongly, though.)

koto: I see. That makes sense. The keyword should only control the `Sanitizer.sanitize*` functions, the SanitizerConfig usage in the default policy can be unrestricted, as whoever controls the default policy can just as well _not_ sanitize at all, so we're not increasing the risk here.

koto: Coming to think of it, most of the times one does not need `trusted-types` directive at all, and this would force those authors to add `trusted-types * 'allow-duplicates' 'sanitizer-api'`, which might be an overkill. 

Perhaps we don't actually need the keyword at all? We're guarding on 'script' sinks for the purpose of DOM XSS. If the sanitizer is guaranteed never to produce a scripty output, it's equivalent to calling native DOM APIs (`createElement`, `setAttribute`, `dataset` and such) which we don't guard on. In fact sanitizer is more strict, as also "javascript:URLs" are rejected. Potentially dropping the keyword  however requires that the baseline of "sanitizer API output does not contain JS" is preserved.

// cc @xtofian

shhnjk: I do not like the idea of `'sanitizer-api'` keyword. I would like `sanitizeToTrustedHTML` to only return safe HTML (i.e. if `SanitizerConfig` is unsafe, `sanitizeToTrustedHTML` method will not be exposed or throw an error with `trustedTypes.emptyHTML` as a return value).

I believe that this will benefit a lot of JS frameworks where they don't have control over HTML or response header. If we require `'sanitizer-api'` keyword, then usage of Sanitizer API in frameworks will break all websites that uses the library with Trusted Types.

koto: The invariant to preserve is that (under `require-trusted-types-for 'script'` for all involved documents) the ability to execute Javascript from a string needs a policy (creation of which might be controlled by `trusted-types`). 

If the sanitizer output - under any configuration - can also be generated with regular non-TT-guarded DOM APIs, it's fine to drop the keyword, as the sanitizer does not give any concerning capability - and the sanitizer as such can create `TrustedHTML` instances. It's just like `trustedTypes.emptyHTML - you don't need a policy for it, as it can't cause XSS. I think the current intention for the API is to be inherently secure like this. 

If this changes and it's possible to output XSSy trees (e.g. some configurations would allow `<iframe srcdoc>`), the integration needs a policy or other opt-in mechanism.

mozfreddyb: I don't think we'll allow the sanitizer to produce what you call "XSSy trees". The built-in allow list ought to only be expandable by custom elements (and, if feasible, only those custom elements that _inherit_ from an allowed element)

mozfreddyb: I don't think we'll allow the sanitizer to produce what you call "XSSy trees". The built-in allow list ought to only be expandable by custom elements (and, if feasible, only those custom elements that _inherit_ from an allowed element)

koto: Double like :)

otherdaniel: Hmm. So: They way we're currently spec-ing (and implementing) Sanitizer the user cannot loosen the built-in defaults, and furthermore the defaults - while not defined yet - should cover all the XSS sources that Trusted Types covers as well. (I think everyone agrees on this.)

The way I look at TT is that is gives you control over the string-to-DOM transition. In that case, an on/off switch in CSP for Santizer-to-TrustedHTML would make a lot of sense. I like koto's definition of "the invariant to preserve is that [...]  the ability to execute Javascript from a string needs a policy". If we agree on that, the on/off switch is not required. In fact, it wouldn't make sense.

Do we agree on that? (If so, I'll just drop the CSP bits. I think so far I'm the only hold-out.)

Also: If we're there: Do we need any integration with TT at all?

We can add the sanitizeToTrustedHTML method, but if santizer (to DocumentFragment) is not-TT checked because it's inherently XSS-safe, then this doesn't buy us much. In any case, I think we can drop both the CSP bits; the TT check; and the policy-from-SanitzerConfig integration. (The last bit was meant to make it easier to use Sanitizer within TT, but if any Sanitizer can generate TT instances directly, that's just an extra indirection.)

koto: > The way I look at TT is that is gives you control over the string-to-DOM transition

Not really. It guards string => XSS sinks. You can create DOM from strings without TT policies already, that's expected. What you cannot create is XSSy DOM. The only reason why we're guarding HTML sinks is because the resulting tree could execute JS.

> We can add the sanitizeToTrustedHTML method, but if santizer (to DocumentFragment) is not-TT checked because it's inherently XSS-safe, then this doesn't buy us much. 

It actually does buy as a lot. It's a terrific way of obtaining safe serializations to pass to HTML sinks, even under TT restrictions. TT + a non-policy way of creating safe HTML allows you to have real-world  applications that work under `require-trusted-types-for 'script'; trusted-types 'none'`, in other words - there is NO code to review, and no JS to inject to every page for security. Everything is guaranteed to be XSS free (modulo some cross-document concerns). Without `sanitizeToTrustedHTML` you can only have that *if* you're not using `innerHTML`, so practically never.

> Drop policy-from-SanitzerConfig integration. 

It potentially allows for optimization, as you'd only have a single, non-exposed instance used internally per-policy.  But I agree it's not a huge advantage.

otherdaniel:
Based on the feedback here, I updated the PR.

- I removed the Trusted Types checks, as well as the `'sanitizer-api'` on/off switch in CSP.
- I removed the sanitizer-in-TT-policy thing and replaced it with an issue of whether this might be needed.

I also added a formalization for `sanitizeToTrustedHTML`. It's rather trivial, though.

index.bs
```diff
@@ -290,3 +290,34 @@ Acknowledgements {#ack}
...
+
+Issue(WICG/sanitizer-api#20): The best way to integrate Trusted Types is still under discussion.
+
+[Trusted Types](https://w3c.github.io/webappsec-trusted-types/dist/spec/) is
+a proposed feature to allow locking down the DOM to prevent cross-site
+scripting attacks. Trusted Types is complementary to the mechanisms proposed
+here. User Agents that support both, this Sanitizer API and Trusted Types,
+benefit from integration between these features.
+
+To allow easy combination of Sanitizer and Trusted Types, the `Sanitizer`
```
shhnjk, line 305: When will `sanitizeToTrustedHTML` not be exposed or throw an error? Do we always want to expose `sanitizeToTrustedHTML` because we will only allow developers to create a `Sanitizer` with a safe tags and attributes? If so, I think that's a bad idea to not to allow customizability of `Sanitizer` to contain a unknow attributes in a safe list.

My idea of integration with Trusted Types is:
1. `sanitizeToTrustedHTML` will not be exposed or throw an error when the `SanitizerConfig` doesn't match the default known-safe `SanitizerConfig`.
2.  Developers can still use `sanitize` or `sanitizeToString`  with unsafe config, but they would need to use `TrustedTypePolicy` for that (so the input to `sanitize` needs to be `TrustedHTML` in this case).

Allowing unsafe config in Sanitizer API seems important. Because we want the default `SanitizerConfig` to be always safe, we don't want to allow unknow attributes such as `data-x`. However, Developer would want to allow that depending on the thier usecase, if they know that it's safe. Allowing them to use Sanitizer API with their scenario seems like an important scenario to support, otherwise I imagine we won't get enough adoption of this API.

koto: What do you mean by "safe"? Both Sanitizer and Trusted Types do not recognize gadgety payloads, as they are application specific. For example, `class` attribute is a gadgety sink in AngularJS, and yet TT will let `div.class` be set with a string, and Sanitizer will let developers to set it - either by default, or by specifying the `SanitizerConfig`. The baseline is "raw" XSS.

If TT gets mechanics to enforce on gadgety sinks (see https://github.com/w3c/webappsec-trusted-types/issues/288 for example), then this should be aligned (i.e. a lax, gadgety sanitizer config should not be a bypass vector for TT configured for gadgets). But we're not there yet. So far the baseline is XSS, so every `SanitizerConfig` should be OK to produce `TrustedHTML`. Am I missing something?

otherdaniel: I think we're discussing several things at the same time: What is the default-safe config; to what extend can it be overridden; and how does Trusted Types fit in.

Default-safe config: I think there's consensus that the defaults should be safe, but I think without an explicit defintion of what that entails. My understanding was that these would be anything that directly executes script (e.g. <script> text content, event handlers), or things that load scripts (e.g. <script src=..>). This would not include unknown elements. IOW, the default config permits anything unless it's in the default known-bad list.

Config overrides: The way I've currently spec-ed it, the config cannot override the defaults, ever. I initially thought differently, but thought I had heard consensus on this. This does presume that the defaults will be a fairly minimal known-bad lists, because otherwise we'd be restricting the applicability of the sanitizer.

I'm happy with both of those, but I now wonder whether we really have consensus on them.

TT integration: I think this will look different depending on the previous issues. In terms of mechanics, Making the TT check depending on the config makes me a bit unhappy. I view TT as effectively part of a method's signature, and changing the signature depending on implied object state is difficult to explain (and to implement).

------------ 

Here's a use case, and I wonder what people think of it:

Suppose, an app decides that the class= attribute is part of their threat model (to pick up koto's example; but could also be id=, or whatever). The app would like to use TT + Sanitzer to guard against this.

- I presume we can't block class (id, etc.) as part of the Sanitizer default, since it's commonly useful and not inherently an XSS source.
- The app can easily configure sanitizers to accomodate this. (`blockAttributes = { "class": ["*"] }`, or so.) But it cannot force all sanitizers to do so, since there's no mechanism to limit sanitizer creation.
- If every Sanitizer can create TrustedHTML, then it becomes difficult to ascertain that all TrustedHTML instances will be "safe" according to the app's chosen criteria.

What do? Should there be an on/off switch for sanitizeToTrustedHTML after all? Can we fix this on either TT or Sanitizer side? Or is this not a realistic use case?

koto: Trusted Types with their `'script'` sink group (the only one existing for now) only enforce on platform XSS sinks. Additional sink support may be added later, but even with this I don't think it would be feasible to allow for re-configuring the semantics of existing platform sinks (like `className` and `class` attribute). The only feasible way is to introduce new sinks (w3c/webappsec-trusted-types#288), maybe even new types for them (https://gist.github.com/koto/1d044f6029ee337beffb4487b80f8b02). 

If authors in the future want to enforce in their custom sinks (and some of those sinks require `TrustedHTML`) then I think `sanitizeToTrustedHTML` would just unconditionallly throw, and these authors would then have to create TT policies that use their tailored sanitizer policies.

otherdaniel: I don't think this is about new sinks. (Other than Sanitzer itself arguably being one.) If an app handles untrusted input, surely it'd make sense to use Sanitzer + TT to _also_ ensure that additional, app-specific critera are met, like no class attributes in formatted input? 

(I agree TT couldn't safe authors from themselves if elsewhere they assign untrusted input to class attributes directly.)

shhnjk: >What do you mean by "safe"? Both Sanitizer and Trusted Types do not recognize gadgety payloads, as they are application specific.

What I think of safe here is only platform specific sinks. If `class` attribute is a gadgety sink in AngularJS, that doesn't matter here, because if TT is enabled, it will throw an TT violation when the gadgety sink will actually hit the platform specific sinks.

>Default-safe config: I think there's consensus that the defaults should be safe, but I think without an explicit defintion of what that entails. My understanding was that these would be anything that directly executes script (e.g. <script> text content, event handlers), or things that load scripts (e.g. <script src=..>). This would not include unknown elements. IOW, the default config permits anything unless it's in the default known-bad list.

I think a Sanitizer with know-bad list is a bad idea, because I've seen many of them failing in the sanitizer space (e.g. WAFs, XSS Auditors/filters, etc) and generally allow list is usually the safer, logical, and practical option. My opinion for Default-safe configs are:

1. Multiple default configs (with all known-good list),  where the weakest default config would be equivalent of DOMPurify (i.e. allow style tag injection). And the weakest config will also be equivalent to TT enforcement (but we will not allow JavaScript URL to links or frames).
2. People can create their own custom Sanitizer with whatever config they want. However, that won't generate a `TrustedHTML`, and therefore they would need to create a Trusted Type policy if they want the output to be a `TrustedHTML`.


Also, I think it's better to discuss these important things in a meeting like WebAppSec, so that everyone is on the same page. I didn't know about most of "consensus" that are being talked here.

koto: I don't think option 2 is on the table. The sanitizer API guarantees that no configuration can become XSSy, so every configuration is safe to generate `TrustedHTML`. How this is actually achieved (e.g. whether the config is based on block- or allowlists) should not be relevant for the TT integration. 

+1 to webappsec, I think we need to discuss the anti-xss approach for the platform anyway, as the sanitizer and CSP only is in my opinion not sufficient.

shhnjk: >I don't think option 2 is on the table.

Yeah, if that's not on the table, then that's fine for the TT integration 😊 But I'm certain that people will come to Sanitizer API and beg to allow `javascript:void(0)` and all sort of compatibility problems 😊

otherdaniel:
I don't think we reached consensus here, so I've tried again with a somewhat different approach in #55. Wdyt?

PR #50, opened 2020-11-26
Specify InputOptions.

otherdaniel:
This specifies SantiizerInput, largely based on #22.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -21,7 +21,12 @@ Markup Shorthands: css off, markdown on
 spec: DOM-PARSING; urlPrefix: https://w3c.github.io/DOM-Parsing/
   type: attribute
     text: innerHTML; for: Element; url: #widl-Element-innerHTML
+  type: method
+    text: parseFromString; url: #widl-DOMParser-parseFromString-Document-DOMString-str-SupportedType-type
 text: window.toStaticHTML(); type: method; url: https://msdn.microsoft.com/en-us/library/cc848922(v=vs.85).aspx
+text: createDocumentFragment; type: method; url: https://dom.spec.whatwg.org/#dom-document-createdocumentfragment
+text: Document; type: interface; url: https://dom.spec.whatwg.org/#interface-Document
+text: DocumentFragment; type: interface; url: https://dom.spec.whatwg.org/#documentfragment
 </pre>
 <pre class="biblio">
 {
@@ -122,8 +127,8 @@ handle additional, application-specific use cases.
     SecureContext
   ] interface Sanitizer {
     constructor(optional SanitizerConfig config = {});
-    DOMString sanitizeToString(DOMString input);
-    DocumentFragment sanitize(DOMString input);
+    DocumentFragment sanitize(SanitizerInput input);
+    DOMString sanitizeToString(SanitizerInput input);
   };
 </pre>
 
@@ -137,6 +142,18 @@ Example:
   // Replace an element's content from unsanitized input:
   element.replaceChildren(new Sanitizer().sanitize(userControlledInput));
 ```
+
+## Input Types {#inputs}
+
+The sanitization methods support three input types: `DOMString`, `Document`,
+and `DocumentFragment`. In all cases, the sanitization will work on a
+{{DocumentFragment}} internally, but the work-fragment will be created by
+parsing, cloning, or using the fragment as-is, respectively.
+
+<pre class="idl">
+  typedef (DOMString or DocumentFragment or Document) SanitizerInput;
+</pre>
+
 
 ## The Configuration Dictionary {#config}
 
@@ -192,7 +209,7 @@ Examples:
   new Sanitizer().sanitizeToString(sample);  // Will output sample unmodified.
 
   // Scripts will be blocked: "abc alert(1) def"
-  new Sanitizer().sanitzeToString("abc <script>alert(1)</script> def");
+  new Sanitizer().sanitizeToString("abc <script>alert(1)</script> def");
 ```
 
 ## Algorithms {#algorithms}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -122,8 +127,8 @@ handle additional, application-specific use cases.
     SecureContext
   ] interface Sanitizer {
     constructor(optional SanitizerConfig config = {});
-    DOMString sanitizeToString(DOMString input);
-    DocumentFragment sanitize(DOMString input);
+    DocumentFragment sanitize(SanitizerInput input);
```
iVanlIsh, line 130: SanitizerInputWithTrustedHTML?

otherdaniel: I suspect not. I'm trying to handle Trusted Types integration in #49, which indeed changes this method to require TrustedHTML.

The problem I'm having is that TT is not presently an accepted standard and I think Sanitzer needs to be usable without TT. WIth 'usable' I mean not only callable from JS without TT, but also specified without relying on the TT spec draft. That is a bit awkward, honestly, but I'm not seeing any good way around this.

mozfreddyb:
@otherdaniel I'm used to single, functional commits that can stand on their own. I feel like I'm getting mixed messages when I'm asked for review but the commits are so tiny and with inconclusive commit messages.
I'm of course totally OK with more smaller changes in tiny commits, but I'd really like their commit message to be more verbose then :-) 

I suppose we can also deal with it here by squashing them together during merge. The effective change isn't too noisy for a review, after all. Just didn't want to leave this unmentioned. :)

otherdaniel:
Hah. I guess I'm used to different tooling here. The intent is to "squash and merge" the commit, so that the messages in the actual commit will disappear, which... is why I put zero effort into having meaningful messages there. The individual commits are effectively my edit history, and can't stand on their own.

I'll squash the branch and will re-upload.

iVanlIsh (approved):
LGTM

index.bs
```diff
@@ -235,8 +252,26 @@ To <dfn>sanitize an attribute</dfn> named |attr|, run these steps:
 6. if |config| has a non-empty [=attribute allow list=] and |name| is not in |config|'s [=attribute allow list=] return 'drop'
 7. return 'keep'
 
+To <dfn>create a document fragment</dfn>
+named |fragment| from a Sanitizer |input|, run these steps:
+
+1. Switch based on |input|'s type:
+  1. if |input| is of type {{DocumentFragment}}, return |input|.
```
mozfreddyb, line 259: I think we need to create a new `DocumentFragment`, in order not to copy potentially assigned observers / event listeners

index.bs
```diff
@@ -235,8 +252,26 @@ To <dfn>sanitize an attribute</dfn> named |attr|, run these steps:
...
+  2. if |input| is of type {{Document}}
+    1. let `f` be the result of {{createDocumentFragment}}.
+    2. return the result of `input.cloneNode(input, true)`.
+  3. if |input| is of type `DOMString`, then:
```
mozfreddyb, line 263: Why parse into document *and* transform into a document fragment?

mozfreddyb (commented):
This is great stuff.

Originally I had the algorithm in my head along the lines below here, but I don't think that's going to work.
I had it string-based, mostly due to the fact that we can't just copy existing an DocumentFragment (observers?, event listeners?).


Gecko allows a few short cuts to pare a string directly into a document fragment, but I'm surprised there's nothing in https://dom.spec.whatwg.org/#interface-documentfragmentor even https://html.spec.whatwg.org/multipage/scripting.html#the-template-element to do that, so that's a bummer :) ...and I think that answers my question above (where I asked why you parse into document and then into fragment..)

```
1. Set |innerHTML| to the empty string
2. Switch based on |input|'s type:
  1. if |input| is of type `DOMString`, then assign |input| to |innerHTML|.
  2. if |input| is of type {{DocumentFragment}}
    1. Serialize it's node tree into a string and assign to |innerHTML|.
  3. if |input| is of type {{Document}}
    1. let |docElement| be the Document's |documentElement| property.
    2. serialize |docElement|'s into a string and assign to |innerHTML|.
3. Create a new {{DocumentFragment}}, whose host is the 
4. Invoke {{createContextualFragment}} parameters |context|, |innerHTML|,

File: index.bs (hunk 2 above)

File: index.bs (hunk 3 above)
Comment (otherdaniel) [reply index.bs:264]: As you guessed: I tried to find a parse-into-DocumentFragment algorithm in the spec, but didn't find any. Hence, the two-step process. I think our implementation can also do that directly; I just didn't find a clean way to spec it.

(Generally, I'm having a hard time finding good references for spec purposes. I could very well be overlooking something here. My understanding is that, as long as the behaviour is unambiguous we're good to go, and the implementations would certainly be free to pick simpler algorithms.)

Comment (otherdaniel) [reply index.bs:264]: As I read it, cloneNode is conveniently specified in a way to only copy the structure.

**Comment by otherdaniel** [comment]
_"I had it string-based, mostly due to the fact that we can't just copy existing an DocumentFragment (observers?, event listeners?)."_   I think [cloneNode](https://dom.spec.whatwg.org/#concept-node-clone) gives us a good option. I'm guesing string-based would also work, except for custom elements & shadow roots. `cloneNode`says _"let copy be a node that implements the same interfaces as node"_, which I'd hope handles those cases.

**Comment by mozfreddyb** [comment]
Right-o. cloneNode it is. :+1:

File: index.bs [hunk 4]
```diff
@@ -235,8 +252,26 @@ To <dfn>sanitize an attribute</dfn> named |attr|, run these steps:
...
+  1. if |input| is of type {{DocumentFragment}}, then:
+    1. let |node| refer to |input|.
+  2. if |input| is of type {{Document}}, then:
+    1. let |node| refer to |input|'s `documentElement`.
+  3. if |input| is of type `DOMString`, then:
+    1. let |node| be the result of the {{parseFromString}} algorithm
+        with |input| as first parameter (`string`),
+        and `"text/html"` as second parameter (`type`).
+2. let `f` be the result of {{createDocumentFragment}}.
+3. return the result of `f.cloneNode(node, true)`.
```
mozfreddyb, line 268: I see `node . cloneNode([deep = false]) ` in spec, but you seem to be using `targetNode.cloneNode(sourceNode, true)` - What am I missing?

otherdaniel: I think I confused myself with importNode and cloneNode. But importNode is Document-only.

 think I now formalized this with correct references.

index.bs
mozfreddyb: We're still using the input fragment here, aren't we?
Rest looks good btw

otherdaniel: I tried to fix that in the 3rd commit, which always to |node|, and then always does a cloneNode of |node| before returning it.

(Should I squash those commits before review? I'm still a little confused about the workflow here.)

otherdaniel: Actuallly.... :)  my version is broken. I still run cloneNode on |input|, rather than |node|. Let me fix that...

mozfreddyb (approved):
:ok_hand:

otherdaniel:
Hrmpf. So... now I didn't squash the commits after all.:-((

Is it okay if I just fix that locally and push the fixed version to the repo, or does that cause issues for other users?

mozfreddyb:
pls do it in #52, so we get a history of changes

PR #51, opened 2020-11-26
Specify allow/blockAttributes.

### LINKED ISSUE #18: Allow/block list for attributes global, or per element?
It seems most sanitizers specify attributes to be alloed/blocked globally (e.g., always remove `onclick=...`). Others have suggested that this is ineffective (or leads to overblocking), and that attributes need to be specifies relative to the elements that hold them.

otherdaniel:
This specifies handling of allow/blockAttributes, following discussion in #18.

I don't think we explcitily agreed on how to handle attributes in the SanitizerConfig. The goal here is to resolve #18, and specifically to
- allow(block specific element/attribute combinations. (E.g. <script src=.>, but not \)
- allow emulation of existing sanitizers (e.g. DOMPurify specifies attribtues independent of the element).

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -255,11 +255,6 @@ Examples for attributes and attribute match lists:
   new Sanitizer({blockAttributes: {"id": ["*"]}}).sanitizeToString(sample);
 ```
 
-Note: `allowElements` creates a sanitizer that defaults to dropping elements,
-  while `blockElements` and `dropElements` defaults to keeping unknown
-  elements. Using both types is possible, but is probably of little practical
-  use. The same applies to `allowAttributes` and `dropAttributes`.
-
 ## Algorithms {#algorithms}
 
 To <dfn lt="sanitize document fragment">sanitize a document fragment</dfn> named |fragment| using |sanitizer| run these steps:
--- a/index.bs
+++ b/index.bs
@@ -329,15 +329,17 @@ named |fragment| from a Sanitizer |input|, run these steps:
 4. [=Append=] the node |clone| to the parent |f|.
 5. Return |f|.
 
+
 Issue(WICG/sanitizer-api#42): It's unclear whether we can assume a generic
   context for {{parseFromString}}, or if we need to re-work the API to take
   the insertion context of the created fragment into account.
 
+
 To <dfn>sanitize</dfn> a given |input|, run these steps:
 
-1, run [=create a document fragment=] algorithm on the |input|.
-2, run the [=sanitize document fragment=] algorithm on the resulting fragment,
-3, and return its result.
+1. run [=create a document fragment=] algorithm on the |input|.
+2. run the [=sanitize document fragment=] algorithm on the resulting fragment,
+3. and return its result.
 
 To <dfn>sanitizeToString</dfn> a given |input|, run these steps:
 
--- a/index.bs
+++ b/index.bs
@@ -168,8 +168,8 @@ dictionary which describes modifications to the sanitze operation.
     sequence&lt;DOMString> allowElements;
     sequence&lt;DOMString> blockElements;
     sequence&lt;DOMString> dropElements;
-    sequence&lt;DOMString> allowAttributes;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -164,9 +164,13 @@ dictionary which describes modifications to the sanitze operation.
 :: The <dfn>element drop list</dfn> is a sequence of strings with elements
    that the sanitizer should remove from the input, including its children.
 : allowAttributes
-:: TODO: <dfn>attribute allow list</dfn>
+:: The <dfn>attribute allow list</dfn> is a map of attributes names a list of
+   elements names, describing which attributes should be allowed on which
+   elements. The special "name" `"*"` denotes all elements.
```
mozfreddyb, line 169: Nit: s/elements/attributes (also occurs on line 173)

Do we really want to offer that kind of footgun?

index.bs
```diff
@@ -222,27 +245,36 @@ To <dfn>sanitize an element</dfn> named |element|, run these steps:
 4. if |name| is in |config|'s [=element drop list=] return 'drop'.
 5. if |name| is contained in the built-in [=default element block list=] return 'block'.
 6. if |name| is in |config|'s [=element block list=] return 'block'.
-7. if |config| has a non-empty [=element allow list=] and |name| is not in |config|'s [=element allow list=] return 'block'
-8. return 'keep'
+7. if |config| has a non-empty [=element allow list=] and |name| is not in |config|'s [=element allow list=] return 'block'.
+8. [=list/iterate|for each=] |attr| in |element|'s [=Element/attribute list=]:
+  1. call [=sanitize an attribute=] with |attr|'s name and |element|'s tag name.
```
mozfreddyb, line 250: I _think_ you need to use something like element's "local name".. We also need to think about namespaces here. Maybe just leave a `// TODO` comment for now

mozfreddyb (changes requested):

index.bs
otherdaniel: I think I meant elements, though.

The (main) idea is that one allows/blocks element/attribute combinations. (Like: script.src) But one might also block attributes absolutely (like: *.src). The main reason why I included that last option is because some existing santizers allow that, and I don't want to force people to re-write all their logic when migrating to sanitizer API. It might also make a lot of sense for some "generic" attributes like event handlers. (Although those would be blocked (and hence non-allowable) by the default config anyhow.)

The notation here is: attr -> [ elements ]. So:

- "src": [ "script" ]
- "onclick": [ "*" ]


With that said: I'm not seeing a footgun. (I do see complexity, which might not be warranted. But I'm seeing anything obvious "sharp edge" in which the user could harm themselves here.)

At the very least, I need to clarify the intent, though. :)

index.bs
otherdaniel: Yes, TODO (or Issue) is totally warranted. An alternative might be to use the constructor name (HTMLScriptElement), since that'd help us with custom elements.

iVanlIsh (approved):
LGTM

index.bs

(Used "local name" instead of tag name, and added an Issue: statement that we will need to revisit this.)

index.bs
otherdaniel: Maybe done. :-)

I've re-written this to be clear(er), hopefully, but have not changed the substance of it. Does this make sense?

index.bs
otherdaniel: I'd really like to get this closed, as it's been out for a while. My preference would still be to land this as-is.


I do wonder about alternatives. I think the use case that you want to block  but not other src= attributes is real. But maybe we can make it easier for people that don't want this, and specify that the attribute lists can be either a list/set - in which case they apply universally - or a dictionary - in which case they select element/attribute combos. (Not sure if that's idiomatic IDL, though.)

otherdaniel:
I'll resolve the merge conflicts and will land it then  Last substantial discussion on this PR was in November. I'll be happy to re-work or revert if we have a plan on how else to handle this.

PR #52, opened 2020-12-04
Revert "Specify InputOptions."

otherdaniel:
Reverts WICG/sanitizer-api#50: This should have been squashed pre-merge. Sorry. :)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -256,16 +256,16 @@ To <dfn>create a document fragment</dfn>
 named |fragment| from a Sanitizer |input|, run these steps:
 
 1. Switch based on |input|'s type:
-  1. if |input| is of type {{DocumentFragment}}, then:
-    1. let |node| refer to |input|.
-  2. if |input| is of type {{Document}}, then:
-    1. let |node| refer to |input|'s `documentElement`.
+  1. if |input| is of type {{DocumentFragment}}, return |input|.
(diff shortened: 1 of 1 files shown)


iVanlIsh (approved):

mozfreddyb:
Well spotted. As long as reverting and relanding with squashed commits is a no-op in terms of spec content, I'm happy for you to just go ahead and make the fix yourself. :-)

otherdaniel:
Dropping this.

My intent was to "clean up" the commit history, since now we're having 4 commits of what should have been one. The way reverts work is that we're adding an "inverse" commit, and then the real commit, meaning we'll have 6 commits of what should have been one. That's not an improvement. :-/

What I could do instead is to commit a clean-up branch, and then force-push the master label onto that branch. However, that'd mean re-writing the commit history, which is an arguably risky and oftentimes non-desireable operation. I guess I shouldn't do that. So: Unless someone feels strongly one way or another, I'll just drop this entirely and will make sure to not do it again, but otherwise leave the present situation as-is.)

(Specifically, I could: `$ git branch master fix-pr-50-squash -f && git push master origin -f`. That'd rewrite the commit history as desired. Observe that `$ git diff master..fix-pr-50-squash` is empty.)

mozfreddyb:
let's leave history then. totally not a big deal :)

Issue #17, opened 2020-07-02
How to identify elements to allow/block?


otherdaniel:
The current draft identifies elements and attributes by name + namespace. But [Custom Elements](https://html.spec.whatwg.org/multipage/custom-elements.html) allow to [customize built-in elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-customized-builtin-example) with a different name.

(In Trusted Types, we decided to go by the constructor name of the base element.)

mozfreddyb:
Do we expect Custom Elements to be added individually and explicitly or do you think we require a switch to allow all custom elements?

otherdaniel:
Not sure, honestly.

I guess this is the type of question where I'd like to get user (web dev) feedback, because I'm not entirely sure I fully understand the problem.

For TT, if I remember correctly, the result was that the constructor name of the type that the attribute is defined on is used. So if you have some custom element `bla` that derives from `HTMLScriptElement`, and you call `.src`, then the callback gets `HTMLScriptElement.src`, not `bla.src` or `script.src`. I was kinda hoping that this could be generalized, so that blocking `HTMLScriptElement` would also block custom elements derived from it. But as said above, I'm not really sure if that is what users would expect.

otherdaniel:
I tried to resolve this issue, by specifying that elements get identied by their constructor name, and then - for custom elements - first run the check with their own constructor, then with the lowest DOM-defined constructor in the constructor chain.

For example, my-script (with MyScript which extends HTMLScriptElement) would be blocked if either MyScript or HTMLScriptElement is blocked.

I thought that's a good idea, until I noticed that most "normal" HTML elements (like <b>, <em>, etc.) are all HTMLElement, and thus the config would no longer be able to distinguis those. That doesn't sound useful.

So: I think we should stick to local names (or qnames, or somesuch). The general idea of checking a custom element both on its own and for its parent tag still seems like a good idea to me.

mozfreddyb:
Yeah, let's operate on a local name and have an explicit if-branch for custom elements.

otherdaniel:
[...]

PR #53, opened 2020-12-16
Specify behaviour of custom elements.

otherdaniel:
Formalize behaviour of custom elements by introducing an "allow custom elements" option.

This PR follows discussion in #17.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -170,6 +170,7 @@ dictionary which describes modifications to the sanitze operation.
     sequence&lt;DOMString> dropElements;
     sequence&lt;DOMString> allowAttributes;
     sequence&lt;DOMString> dropAttributes;
+    boolean allowCustomElements;
   };
 </pre>
 
@@ -187,6 +188,12 @@ dictionary which describes modifications to the sanitze operation.
 :: TODO: <dfn>attribute allow list</dfn>
 : dropAttributes
 :: TODO: <dfn>attribute drop list</dfn>
+: allowCustomElements
+:: <dfn>allow custom elements option</dfn> determines whether
+   [=custom elements=] are to be considered. The default is to drop them.
+   If this option is true, custom elements will still be checked against all
+   other built-in or configured configured checks. Note that whether an
+   element is recognized or not depends on the current document.
 
 Note: `allowElements` creates a sanitizer that defaults to dropping elements,
   while `blockElements` and `dropElements` defaults to keeping unknown
@@ -230,9 +237,19 @@ To <dfn lt="sanitize document fragment">sanitize a document fragment</dfn> named
 
 To <dfn>sanitize a node</dfn> named |node| run these steps:
 
-1. if |node| is an element node, call [=sanitize an element=] and return its result.
+1. if |node| is an element node, call [=sanitize an element node=] and return its result.
 2. if |node| is an attribute node, call [=sanitize an attribute=] and return its result.
 3. return 'keep'
+
+To <dfn>sanitize an element node</dfn> |node|, run these steps:
+
+1. Call [=look up a custom element definition=] with the current document,
+   the |element|'s namespace, the |element|'s local name, and the |element|'s
+   `is` attribute (if present).
+2. If the previous step returns anything other than `null`, and if the
+   |sanitizer|'s [=configuration=]'s [=allow custom elements option=] is unset
+   or set to anything other than `true`, return 'drop'.
+3. Call [=sanitize an element=], using |node| as the |element|, and return its result.
 
 To <dfn>sanitize an element</dfn> named |element|, run these steps:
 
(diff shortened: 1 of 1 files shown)


mozfreddyb:
I think it might be easier and more useful to check whether an element that is considered unknown is a potential custom element based on its local name being a [valid custom element name](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name). I.e., custom elements always contain a dash.

I think it would be really convenient if we could allow custom elements regardless of whether they have been registered (yet).

otherdaniel:
> I think it might be easier and more useful to check whether an element that is considered unknown is a potential custom element based on its local name being a valid custom element name. I.e., custom elements always contain a dash.

Done, and indeed easier. This also removes any dependency on the document and registration state.

index.bs
```diff
@@ -238,12 +244,18 @@ To <dfn>sanitize an element</dfn> named |element|, run these steps:
...
-3. if |name| is contained in the built-in [=default element drop list=] return 'drop'.
-4. if |name| is in |config|'s [=element drop list=] return 'drop'.
-5. if |name| is contained in the built-in [=default element block list=] return 'block'.
-6. if |name| is in |config|'s [=element block list=] return 'block'.
-7. if |config| has a non-empty [=element allow list=] and |name| is not in |config|'s [=element allow list=] return 'block'
-8. return 'keep'
+3. if |name| is a [=valid custom element name=] and if |config|'s
+   [=allow custom elements option=] is unset or set to anything other than `true`, return 'drop'.
+4. if |name| is contained in the built-in [=default element drop list=] return 'drop'.
+5. if |name| is in |config|'s [=element drop list=] return 'drop'.
```
mozfreddyb, line 250: I think it would be nice if we could run an initialization algorithm that combines a built-in list with a config list and then we'd just checks against that list during sanitization. Happy to have that happen in a follow up though :)

mozfreddyb (approved):

index.bs
otherdaniel: Hm. Why?

I had thought about this early on and had initially tried to write it that way. But I don't think it's that easy to cleanly spec it that way, and some of the complexity will still remain. E.g., I don't think we can merge the "custom element name" condition into a list. If so, we'd end up with an algorithm that's split between two different places of the spec, which I suspect will be harder to understand since then the reader needs to connect the two parts on their own. So arguably, the current version is a bit clumsy, but it does strike me as easy to understand.

mozfreddyb:
If I understand it correctly from our spec text, we require a custom element to be explicitly in the supplied config AND the `allowCustomElements` option set to true. Why? Can't we just accept supplied custom element names as long as they don't inherit from a known-bad element that's not in our baseline (like script)?

otherdaniel:
> If I understand it correctly from our spec text, we require a custom element to be explicitly in the supplied config AND the allowCustomElements option set to true.

This is how I understand it as well.

> Why? Can't we just accept supplied custom element names as long as they don't inherit from a known-bad element that's not in our baseline (like script)?

The first draft of this PR required a look-up in the custom elements registry, where we could have made such a call, but based on the PR feedback I had changed that to be name based. Maybe that was over-zealous?

My understanding is that the relationship between custom elements and "parent" is in the document registry, and can change over time (as new elements are added). That adds complexity, and it adds risk, if e.g. content is sanitized early on (and the doc will register additional custom elements later), or if sanitized content is re-used in another frame (e.g. via postMessage; or maybe a roundtrip through storage).


That said... I don't mind changing this at all. The only "real" advantage I see of the current rules is that they're easy (to spec, and to explain). I'll be happy to do this elsehow, if we have a sufficiently clear proposal on how else to do it. Do you have specific semantics in mind?

franktopel:
Please keep in mind that explicit white-listing of tag names really is not an option here.

Imagine this scenario (which is what we currently develop):

* Library `Foo` comes with a lot of atomic gui components and ist the very first thing loaded in any application based on it. As such, this is the point in time where we need to define the trusted types policies, and as we don't want to allow additional trusted type policies to be created later on, nothing can be changed or added to the policies later.
* Application `Bar` consumes `Foo`. The application itself is a set of custom elements that serve as view components and view controllers, as well as as an application controller holding general state for the application and controlling which view controllers are instantiated (e.g. the application controller also takes care of routing for the SPA).
* Half a year later the same customer starts a new application `Baz` that follows the same architectural patterns as `Bar` does, and also is based off `Foo`.
* Another 6 months in, the customer starts integrating `Bar` and `Baz` into an "umbrella" application `BazBarAndMore`. This application comes with a `<company-portal>` web component that controls the instantiation of `Bar` and `Baz`.

koto:
From my limited understanding, the most custom elements can do is to enable a script-gadgety behavior e.g. define an attribute which will call into `eval` etc. Custom elements that extend native elements still need their base element in the DOM with `<script is="actually-my-script">` notation, and as such, the XSSy ones would always be discarded. 

Sanitizers would have a hard job protecting against gadgets. Only Trusted Types can address them, since the actual sink call (and the XSS) would be in the definition of the component in JS, and not in the HTML. Since gadgets are not in scope for sanitizer, I think it might be OK (as in - the security baseline still stands) to let the callers specify that they allow all custom elements, as long as the XSSy nodes are still processed, and one can't re-allow script by `<script is="foo-bar">`.

Allowlist mode for custom element names is probably also useful for some callers, so I guess both should be supported via config?

All that said, I didn't work with custom elements much, so I might be wrong on my security assumptions. Calling @domenic.

domenic:
I'm not sure if there's a specific question, but @koto's summary seems correct to me.

koto:
I was just wondering if I missed something and you can make `<foo-bar>`behave like a native script without actually having JS code that will just call into `eval`, `import` or other execution sinks. Thanks for confirming!

PR #54, opened 2021-01-21
Add note about input normalization.

otherdaniel:
This spells out a consequence of using the existing HTML parser machinery, that may not be obvious to someone unfamiliar with the specs.


mozfreddyb (approved):
Very much worth spelling out. Thanks!

PR #60, opened 2021-02-11
Re-write default handling and supply default values.

otherdaniel:
This PR
- distinguishes between built-ins (non-overridable) and defaults (what you get without an explicit config)
- makes the default an allow-list (plus a provisional allow list for attributes)
- I snuck in special-casing for &lt;template>

The first commit defined the built-ins as a block-list; the second also as an allow-list.
I find the second one to be somewhat harder to read. (E.g. it's more difficutl to find the difference between built-in and default lists.)


index.bs
```diff
@@ -169,22 +169,32 @@ Note: Sanitizing a string will use the [=HTML Parser=] to parse the input,
...
 <pre class="idl">
   dictionary SanitizerConfig {
     sequence&lt;DOMString> allowElements;
+    sequence&lt;DOMString> allowContentElements;
     sequence&lt;DOMString> blockElements;
     sequence&lt;DOMString> dropElements;
     AttributeMatchList allowAttributes;
     AttributeMatchList dropAttributes;
+    boolean allowUnknownElements;
+    boolean allowUnknownAttributes;
```
mozfreddyb, line 186: I don't want to turn this into endless bikeshedding, but I think a naming scheme where sequences and boolean flags are both named `allow...` is confusing. I don't have a better suggestion as of yet, but would like to address this somehow.

index.bs
```diff
@@ -198,6 +208,14 @@ dictionary which describes modifications to the sanitze operation.
 : dropAttributes
 :: The <dfn>attribute drop list</dfn>  is an [=attribute match list=], which
    determines whether an attribute (on a given element) should be dropped.
+: allowUnknownElements
```
mozfreddyb, line 211: I don't think we want a blanket allow-boolean for this case (and the attribute case below). I think we should ask developers to specifically list the elements & attributes they need.

I know this puts an extra onus on e.g., framework authors, but I'd rather do this than provide a footgun that turns this  into a security bug. Happy to debate this further.

index.bs
```diff
@@ -359,36 +415,518 @@ To <dfn>sanitizeToString</dfn> a given |input|, run these steps:
...
+Note: The default lists are expected to change when user agents support new
+    elements or attributes, with regard to only those new elements or attributes,
+    without an explicit re-release of this specification.
+
+: Builtin Drop Elements
+
+:: The <dfn>built-in element allow list</dfn> has the following value:
+    ```
+    [
+          "a", "abbr", "acronym", "address", "area", "article", "aside", "audio",
```
mozfreddyb, line 437: Nit: For the sake of "git blame" and maintenance, is there an easy way to list them line-by-line? Maybe we can use [bikeshed's include file feature](https://tabatkins.github.io/bikeshed/#including)?

index.bs
```diff
@@ -359,36 +415,518 @@ To <dfn>sanitizeToString</dfn> a given |input|, run these steps:
...
+
+: Default Configuration
+
+:: The <dfn>default configuration</dfn> has the following value:
+    ```
+    {
+      "allowUnknownElements": false
+      "allowUnknownAttributes": false,
+      "allowCustomElements": false,
+      "allowElements": [
```
mozfreddyb, line 689: Same nit as above, plus I'd prefer we figure out a way to name it once in the spec and then reference it from there instead of having it fully enumerated in two places.

index.bs
otherdaniel: Done. This is indeed much nicer.

index.bs
otherdaniel: Not sure I agree. The baseline and the default config will differ in subtle ways. Otherwise, we could just merge them again.

Arguably, it'd make sense to have an easier way to construct them, like:
default = { allowElement: baseline.allowElements + ["bla", "bla", "blubb"] }
Or so. That'd likely be much easier to understand.

Incidentally, today someone brought up exactly that point when discussing allowUnknownElements. So if someone wants to add an "unknown" element to the list, they shouldn't have to research and repeat the entire allow-list. I guess we're missing a mechanism there.

(Could be: Default config exposed, so people can easily build on it. Or every Sanitizer exposes their config. Or "inheritance", where you name a "parent" Sanitizer. Or... )

index.bs
otherdaniel: I'm not sure I agree, honestly. I'll remove them in the next update, but I guess I'll make an attempt at getting them back in again. :)

index.bs
otherdaniel: Hmm. I can't come up with a super good name. If we remove the allowUnknown* options, the issue becomes moot, though.

resources/baseline-config.json
```diff
@@ -0,0 +1,755 @@
+{
```
otherdaniel, line 1: I've run these files through a tool called json_pp (JSON pretty print) that I found on my machine. I'm not sure this format is the best, but it's very diff-able.

index.bs
```diff
@@ -285,17 +293,29 @@ To <dfn>sanitize an element</dfn> named |element|, run these steps:
...
+   Note: The sanitizer should not actually assign any interfaces to the elements
+       it processes. We're just using the HTML concept to define whether an
+       element is unknown.
+4. if |kind| is `regular` and if |name| is not contained in the
+   [=baseline configuration=]'s [=element allow list=], return 'drop'
+5. if |kind| is `custom` and if |config|'s [=allow custom elements option=] is
+   unset or set to anything other than `true`, return 'drop'.
+7. if |name| is in |config|'s [=element drop list=] return 'drop'.
+8. if |name| is in |config|'s [=element block list=] return 'block'.
+9. if |config| has a non-empty [=element allow list=] and |name| is not in |config|'s [=element allow list=] return 'block'
```
iVanlIsh, line 313: element allow list seems to be always non-empty since we have a default list.

index.bs
otherdaniel: Would it make sense to add a (non-normative, I guess?) section that explains how the values were constructed? Ideally in executable form.

Something like: https://github.com/otherdaniel/purification/blob/default-construction/resources/configurations.html

The idea is that one can copy&paste the results into the spec (or generate them as part of the deploy script). If all goes well, the code version could be more easily readable & modifyable than the spec version, but we'd still have a definite version in the spec.

otherdaniel:
I've uploaded a re-write of this PR. I'll re-add actual default values next.
@mozfreddyb could you take a quick look and offer an opinion on whether this is going in the right direction?

otherdaniel:
I've also (re-) added actual default values, as well as a script that computes them. The intention is that the script is easier to understand and modify, but we'd still have exact values in the spec. I'd appreciate another round of reviews, looking both at the actual default values, as well as the mechanics for how we construct them.

mozfreddyb (commented):
Thanks for the hard work! Sorry it took me so long.

index.bs
```diff
@@ -337,59 +296,136 @@ named |fragment| from a Sanitizer |input|, run these steps:
...
+helper definition:  The <dfn>element kind</dfn> of an |element| is one of
+`regular`, `unknown`, or `custom`. Let |kind| be:
+- `custom`, if |element|'s tag name is a [=valid custom element name=],
+- `unknown`, if |element| is not in the [[HTML]] namespace or if |element|'s
+   tag name denotes an unknown element &mdash; that is, if the
+   [=element interface=] the [[HTML]] specification assigns to it would
+   be {{HTMLUnknownElement}},
+- `regular`, otherwise.
+
+Similarly, the <dfn>attribute kind</dfn> of an |attribute| is one of `regular`
```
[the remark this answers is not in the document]
otherdaniel: As I read it, `<data>` is `regular` (because of the HTMLDataElement interface), `<data-foo>` would be `custom`.

I think HTML spec doesn't have `<data-*>` elements , but it does have `data-*=` attributes. But that means we just have the same problem just a few lines down. :)  I'd think that `data-foo=` would also be `regular`, although that isn't as nicely defined.

Note that in this PR I've dropped the idea of an `allow-unknown:` option, so there's a blanket allow-all option only for custom elements, not for unknowns.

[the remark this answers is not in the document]
mozfreddyb: Sorry, this was specifically meant for.

index.bs
```diff
@@ -168,8 +168,8 @@ Note: Sanitizing a string will use the [=HTML Parser=] to parse the input,
 
 ## The Configuration Dictionary {#config}
 
-The <dfn lt="configuration">sanitizer's configuration object</dfn> is a
-dictionary which describes modifications to the sanitze operation.
+The sanitizer's <dfn>configuration object</dfn> is a
+dictionary which describes modifications to the sanitize operation.
```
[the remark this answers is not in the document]
otherdaniel: I'm not sure if this is an editorial comment, or a comment on how the defaults work.

Editorial: The config certainly does describe a modification to the sanitize operation. 

Defaults: The way I've specified it here is that the default is a proper default, and a user supplied config replaces the default. (As opposed to the baseline, which is always enforced.) I've now changed it to be more granular, so that the default allow lists is used whenever no allow list is supplied. That is, a user config with a block list would apply the user's block list + the default allow list. Is this what you meant?

index.bs
```diff
@@ -337,59 +296,136 @@ named |fragment| from a Sanitizer |input|, run these steps:
...
+
+Note: This definition of stricter actions ensures that the built-in baseline
+      configuration cannot be overriden, and therefor forms a hard guarantee
+      for all Sanitizer instances.
+
+Before describing how an effective configuration is derived, we need a
+helper definition:  The <dfn>element kind</dfn> of an |element| is one of
+`regular`, `unknown`, or `custom`. Let |kind| be:
+- `custom`, if |element|'s tag name is a [=valid custom element name=],
+- `unknown`, if |element| is not in the [[HTML]] namespace or if |element|'s
```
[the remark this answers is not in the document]
otherdaniel: I'd prefer not to. (yet.)

Currently, this doesn't cover any non-HTML namespace. (All non-HTML is `unknown`, thus blocked by default, but overridable.) I think we need to do a second pass, where we have proper defaults for SVG + MathML, and then extend the spec to handle those cleanly. But as long as those parts aren't in place we arguably shouldn't pretned to handle any non-HTML.

I've added an issue to the spec, so this doesn't get lost.

mozfreddyb:
oh darn. merging #70 before this made this a revision with merge conflicts. :confounded:

mozfreddyb:
let's land this :)

otherdaniel:
Landed! :)

PR #61, opened 2021-02-12
Add FAQ.

otherdaniel:
The intro/README page contains an FAQ which is very potato. Let's make it more Croquette.

This adds an FAQ as a separate page.  The goal is to replace the intro page FAQ.

The questions here focus on the API user, and try to ask the questions that I've actually seen coming up. (E.g. questions like, 'are the defaults safe?') I think the original FAQ was less targets at developers, and more targeted at browser strategists to whom we wanted to explain why this API is even needed. Now that we can see where the
Santizer is going, I think we can change focus to developers.

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -22,7 +22,9 @@ We expect to be working through these issues in the coming months.
 
 If you wish to try it out already:
 
-* Firefox: Enable the `sanitizer.enabled` preference.
+* Firefox: Go to `about:config`, search for the `dom.security.sanitizer.enabled`
+  flag and set it to true
+
 * Chromium / Chrome: Start the browser with the
   `--enable-blink-features=SanitizerAPI` flag.
 
@@ -38,8 +40,9 @@ const default_sanitizer = new Sanitizer();
 // Returns a DOM Fragment equivalent to "<em>Hello World</em>".
 default_sanitizer.sanitize("<em>Hello World</em>");
 
-// Complex formatting is allowed by default. This will pass unchanged.
-default_sanitizer.sanitize("<div id='hello'><span style='font-weight: bold'><Hello World</span></div>");
+// Complex formatting is allowed by default. This will return a fragment
+// equivalent to its input.
+default_sanitizer.sanitize("<div id='hello'><span style='font-weight: bold'>Hello World</span></div>");
 
 // Script execution is no-no: Returns "<p>Hello</p>"
 default_sanitizer.sanitize("<p onclick='alert(\'surprise\')'>Hello</p>");
@@ -53,7 +56,7 @@ overridden by custom configurations.
 
 ```js
 // Produces an empty fragment. There is no way to allow the <script> element.
-new Sanitizer({"allowElements": ["script"]}.sanitize("<script>alert('surprise')</script>");
+new Sanitizer({"allowElements": ["script"]}).sanitize("<script>alert('surprise')</script>");
 ```
 
 Other notions of safety can likely be configured. For example, if an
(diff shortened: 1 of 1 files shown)


otherdaniel:
The formatted view is easier to read: https://github.com/otherdaniel/purification/blob/faq/faq.md

faq.md
```diff
@@ -0,0 +1,99 @@
...
+
+Not yet. Current browser support is preliminary and incomplete and still
+diverges in several details. And the specification itself is still in flux.
+We expect to be working through these issues in the coming months.
+
+If you wish to try it out already:
+
+* Firefox: Enable the `sanitizer.enabled` preference.
+* Chromium / Chrome: Start the browser with the
+  `--enable-blink-features=SanitizerAPI` flag.
```
mikewest, line 27: Can we put it behind chrome://flags/#enable-experimental-web-platform-features?

faq.md
```diff
@@ -0,0 +1,99 @@
...
+It is safe against any direct XSS.
+
+```js
+const default_sanitizer = new Sanitizer();
+
+// Returns a DOM Fragment equivalent to "<em>Hello World</em>".
+default_sanitizer.sanitize("<em>Hello World</em>");
+
+// Complex formatting is allowed by default. This will pass unchanged.
+default_sanitizer.sanitize("<div id='hello'><span style='font-weight: bold'><Hello World</span></div>");
```
mikewest, line 42: Is the `<` before "Hello" a typo, or an example? If it's an example, I'm not sure "pass unchanged" is a good description?

faq.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Is Sanitizer output secure?
+
+'Secure' means different things in different contexts. The Sanitizer guarantees
+safety from direct script execution in all configurations. This cannot be
+overridden by custom configurations.
+
+```js
+// Produces an empty fragment. There is no way to allow the <script> element.
+new Sanitizer({"allowElements": ["script"]}.sanitize("<script>alert('surprise')</script>");
```
mikewest, line 56:
```suggestion
new Sanitizer({"allowElements": ["script"]}).sanitize("<script>alert('surprise')</script>");
```

faq.md
```diff
@@ -0,0 +1,99 @@
...
+default_sanitizer.sanitize("<div id='hello'><span style='font-weight: bold'><Hello World</span></div>");
+
+// Script execution is no-no: Returns "<p>Hello</p>"
+default_sanitizer.sanitize("<p onclick='alert(\'surprise\')'>Hello</p>");
+```
+
+## Is Sanitizer output secure?
+
+'Secure' means different things in different contexts. The Sanitizer guarantees
+safety from direct script execution in all configurations. This cannot be
```
mikewest, line 51: Might be worth explaining "direct"?

faq.md
```diff
@@ -0,0 +1,99 @@
...
+
+Other notions of safety can likely be configured. For example, if an
+application relies on the `id=` attribute and does not wish user-supplied input
+to introduce unforseen `id=` attribute values, then the application can easily
+configure a block a Sanitizer to block the this attribute. However, the
+application must ensure on its own that all parts of the application
+(including any of its dependencies) will adhere to this restriction.
+
+
+## When should I use `.sanitize`, when `.sanitizeToString`?
```
mikewest, line 67: Maybe the following is clearer?

```suggestion
## When should I use `.sanitize` vs `.sanitizeToString`?
```

faq.md
```diff
@@ -0,0 +1,99 @@
...
+
+This is up to you. `.sanitize` and `.sanitizeToString` do the same work,
+except `.sanitizeToString` will serialize the result fragment into a string
+and return that. Since in all likelihood the Sanitizer result will eventually
+be inserted into the DOM, retaining the fragment (i.e. `.sanitize`) rather
+than a serialization that needs to be re-parsed (i.e. `.sanitizeToString`)
+is more efficient.
+
+Additionally, handling a string opens up the possibility of mXSS if the
```
mikewest, line 76: Maybe link "mXSS" off to something like https://hackinparis.com/data/slides/2013/slidesmarioheiderich.pdf?

mikewest: Ah, you had a TODO below. :)

faq.md
```diff
@@ -0,0 +1,99 @@
...
+Yes, please. We see these as APIs that solve different aspects of the
+same problem. THey are separate, but should work well together.
+
+Details of Santizer API/Trusted Types integration are still being worked out.
+
+
+## Can I sanitize myself rich?
+
+Of course. Firefox's Client Bug Bounty Program includes the Sanitizer API
+under certain conditions. See: https://twitter.com/freddyb/status/1304331985398235138
```
mikewest, line 98: Presumably we'd include it in Chrome's VRP as well?

mikewest (approved):
LGTM % nite.

faq.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Can I use the Sanitizer in my app?
+
+Not yet. Current browser support is preliminary and incomplete and still
+diverges in several details. And the specification itself is still in flux.
+We expect to be working through these issues in the coming months.
+
+If you wish to try it out already:
+
+* Firefox: Enable the `sanitizer.enabled` preference.
```
mozfreddyb, line 25: Go to `about:config`, search for the `dom.security.sanitizer.enabled` flag and set it to true

faq.md
otherdaniel: That's the plan, but so far it isn't. And we're not quite ready yet, since our current defaults are terrible. :) 
I don't want to sneak in any Chrome-related announcements, so I think we should only metnion Chrome VRP here once it is actually in.

faq.md
mozfreddyb: :+1:  to what Mike said. We shouldn't promise "unchanged", but either functionally-equivalent or XSS-free.

faq.md
mozfreddyb: I'd nudge towards sanitize over the string-variant more intensively. Specifically, I wouldn't start with "This is up to you".

How about:
```
We're recommending everyone to use `sanitize()` if possible:
* There are XSS risks with parsing & reparsing HTML strings, which are known as mXSS [link].
* In order for the Sanitizer to work, we're already parsing the supplied input into a DocumentFragment and analyze the DOM tree to identify XSS risk.. This DOM tree can be easily applied to an existing Node in your document. Turning it into a string only to just parse it again within the implementation of innerHTML carries a performance cost, that we can help you avoid.```

File: faq.md [hunk 9]
```diff
@@ -0,0 +1,99 @@
...
+and to use `.sanitizeToString` in cases where existing code structure or
+other constraints make it difficult to use a type different from string.
+
+TODO: Link to mXSS explanation.
+
+
+## Can I use the Sanitizer API together with Trusted Types?
+
+Yes, please. We see these as APIs that solve different aspects of the
+same problem. THey are separate, but should work well together.
```
mozfreddyb, line 90: Typo: s/THey/They/

faq.md
mozfreddyb: I like the question phrasing :joy: 
Instead of my tweet, please link to https://www.mozilla.org/en-US/security/client-bug-bounty/#exploit-mitigation-bounty :)

mozfreddyb (commented):
Great, thank you! Left a couple of nits but otherwise super happy :)

faq.md
otherdaniel: Yes, we should. But that's a Chrome feature request, not a Sanitizer FAQ. :-)

faq.md
otherdaniel: Done.

`<` was typo. Also changed the wording.

faq.md
otherdaniel: Done.

faq.md
otherdaniel: Done.

faq.md
otherdaniel: Done.

faq.md
otherdaniel: Hmm. Generally, I'd like to stay in _"I'm-a-sign-not-a-cop"_ territory. I've reworded this along your suggestion, but kept it somewhat more explanatory and less prescriptive. Care to have another look?

otherdaniel: Link: Done.

faq.md
otherdaniel: Done.

faq.md
otherdaniel: Done.

faq.md
otherdaniel: Done.

faq.md
```diff
@@ -63,37 +66,49 @@ configure a block a Sanitizer to block the this attribute. However, the
...
-The recommendation of the Sanitizer API spec editors is to prefer `.sanitize`,
-and to use `.sanitizeToString` in cases where existing code structure or
+Note: When we refer to "direct" script execution, we mean all markup that will
+  cause the browser to parse and execute script, such as the `<script>` element
+  or the `onclick=` event handler attribute. The adjective "direct" is meant
+  to distinguish this from indirect script execution via
+  [script gadgets](https://dl.acm.org/doi/abs/10.1145/3133956.3134091). Script
+  gadgets are application dependent and cannot be sanitized in a generic API.
+  But you can likely use custom Sanitizer configurations adapted to your
+  applications needs.
```
mozfreddyb, line 76: "application's needs?"

faq.md
```diff
@@ -63,37 +66,49 @@ configure a block a Sanitizer to block the this attribute. However, the
...
+
+* Since in all likelihood the Sanitizer result will eventually
+  be inserted into the DOM, retaining the fragment (i.e. `.sanitize`)
+  rather than a serialization that needs to be re-parsed
+  (i.e. `.sanitizeToString`) is more efficient, since you avoid un-parsing
+  and then re-parsing the same data that has already been parsed for us.
+
+The Sanitizer API spec editors recommend to prefer `.sanitize`, and to use
+`.sanitizeToString` in cases where existing code structure or
 other constraints make it difficult to use a type different from string.
```
mozfreddyb, line 100: Excellent, thank you! :)

mozfreddyb (approved):
🎆

faq.md
otherdaniel: Done.

PR #62, opened 2021-02-15
FAQ: Add question about extensibility.

otherdaniel:
Another FAQ entry.

This one is admittedly not frequently asked at all; but I do anticipate that we'll sooner or later get into some sort of feature frenzy, and I'd like to provide at least some sort of answer ahead of time.


mozfreddyb:
+1 to content and perspective, but I'm not sure it's worth adding to the FAQ.

otherdaniel:
Would there be another place where this could go?

My intention is: Everyone I talk to warns me about feature creep for Sanitizers. I think we can head that off, at least by some extent, by providing a clear path of how to build you own stuff on top of it, with maybe an example or two. It clearly doesn't belong into the spec. You're right that it arguably doesn't belong into an FAQ, at least not until it has actually been asked frequently. Where should we put this kind of into?

mozfreddyb:
Ah, let's just put it into the FAQ. :-)

mozfreddyb (approved):

Issue #58, opened 2021-02-05
traverse and sanitize within template element?


mozfreddyb:
Imagine a fragment like
```html
<template><img src="x" onerror="alert(1)"></template>
```

Do we expect the sanitizer to special-case the `<template>` element and traverse into it's `content` attribute?
Firefox currently doesn't support the element, but I'd like to support it and traverse.
From looking at [this testcase](https://github.com/cure53/DOMPurify/blob/98e1ee438798dd11c144c0c5d57648fbfd28fac4/test/fixtures/expect.js#L191) DOMPurify does so too.

From a spec perspective, I think we could check if the element has a `content` attribute of type DocumentFragment (or such), which might also handle Declarative Shadow DOM.

otherdaniel:
I haven't use `<template>`, so I'm not sure if I understand this correctly, but the way I read the docs is that the `.content` attribute mirrors the actual element content, basically like a read-only version of `.innerHTML`. If so, I'd expect no special-casing to be necessary, in that I'd assume Sanitzer would recurse into the children - like for any other element - and `.content` would end up reflecting whatever the surviving content is.

In your example::
  ```<template><img src="x" onerror="alert(1)"></template>  // .content == <img src="x" onerror="alert(1)">```
Post `.sanitize`:
  ```<template><img src="x"></template>  // .content == <img src="x">```

Does DOMPurify acutally special-case `.content`, or is that a fallout of recursing into the children?

mozfreddyb:
Well `content` is not an attribute of the HTML element, but a property of the DOM Node. But I think we agree nonetheless.

jonathanKingston:
Speaking with @mozfreddyb about this;  I think `<template>`, `<slot>` and `data-*` are all of the same ilk, undefined user extensions that can't be guaranteed to be safe in the implementing application.

It's possible with any DOM node to have this risk on injection, but the expectation on these three are clear userland exceptions that a web developer can treat as _here be dragons™_

otherdaniel:
Not sure I follow. 

- `<template>` and `<slot>` seem to contain HTML data. Which is arguably exactly what an HTML sanitizer would know how to deal with.

- `data-*` just appears to be text strings that are easily accessible by CSS or custom elements, but fundamentally the same as any other attribute of text content.

Am I still something?

It does strike me that they should be default-blocked. But blocking them in a non-overridable fashion seems a bit much, and in that case we need to define how to handle them.

jonathanKingston:
[...]

PR #63, opened 2021-02-15
Introduce special-case handling.

otherdaniel:
In repsonse to issue #58, introduce special case handling for certain elements. So far, only `template` is affected.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -325,13 +325,20 @@ expressed in terms of configuration options or other algorithms. The following
 algorithm collects these in one place. To <dfn>handle funky elements</dfn>,
 run these steps:
 
-1. if |element|'s name is `template`:
-  1. Let |content| be value of the `content` attribute of |element|,
-     treated as a {{HTMLTemplateElement}}.
+1. if |element|'s [=element interface=] is {{HTMLTemplateElement}}:
   2. Run the steps of the [=sanitize document fragment=] algorithm on
-     |content|, and replace |element|'s |content| attribute with the result.
+     |element|'s |content| attribute, and replace |element|'s |content|
+     attribute with the result.
   3. Drop all child nodes of |element|.
-
+2. if |element|'s [=element interface=] has a {{HTMLHyperlinkElementUtils}} mixin:
+  1. if |element|'s `protocol` property is "javascript:", then remove the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -319,6 +320,17 @@ To determine whether an |attribute| and |element| <dfn>attribute-match</dfn> an
 4. if |matches| contains the string "*", return true.
 5. return false.
 
+Some HTML elements require special treatment in a way that can't be easily
+expressed in terms of configuration options or other algorithms. The following
+algorithm collects these in one place. To <dfn>handle funky elements</dfn>,
+run these steps:
+
+1. if |element|'s name is `template`:
+  1. run the sanitizer on the `content` attribute as if `sanitizeToString`
```
mozfreddyb, line 329: 1. Can we point to our "sanitize a document fragment" algorithm?
2. The [template's content attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template#attributes) is a DocumentFragment. I think we might be able to replace the content attribute with the newly created DocumentFragment. No string serialization required, afaiu.

otherdaniel: Good idea. Done.

index.bs
```diff
@@ -319,6 +320,26 @@ To determine whether an |attribute| and |element| <dfn>attribute-match</dfn> an
...
+     |element|'s |content| attribute, and replace |element|'s |content|
+     attribute with the result.
+  3. Drop all child nodes of |element|.
+2. if |element|'s [=element interface=] has a {{HTMLHyperlinkElementUtils}} mixin:
+  1. if |element|'s `protocol` property is "javascript:", then remove the
+     `href` attribute from |element|.
+3. if |element|'s [=element interface=] is {{HTMLFormElement}}:
+  1. if |element|'s `action` attribute is a [[URL]] with `javascript:`
+     protocol, remove the `action` attribute.
+4. if |element|'s [=element interface=] is {{HTMLInputElement}}:
```
mozfreddyb, line 339: nit: formaction attribute also exists on HTMLButtonElement

mozfreddyb (approved):

index.bs
otherdaniel: Yes, thank you. Added.

PR #65, opened 2021-02-23
Initial text for Security Considerations

mozfreddyb:
This is the initial text, I envision to discuss the problems we are and aren't solving with this spec.

I'm honestly not sure whether we need references for things like mXSS, DOM clobbering and Script Gadgets, as this is non-normative, but can add if necessary.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -393,7 +393,7 @@ by a custom configuration.
 # Security Considerations {#security-considerations}
 
 The Sanitizer API is intended to prevent DOM-Based Cross-Site Scripting
-by traversing a DocumentFragment and removing elements and attributes
+by traversing a supplied HTML content and removing elements and attributes
 according to a configuration. The specified API must not support
 the construction of a Sanitizer object that leaves script-capable markup in
 and doing so would be a bug in the threat model.
@@ -420,24 +420,22 @@ properties like `children` of an HTML element in the DOM are overshadowed by
 the malicious content.
 
 The Sanitizer API does not protect DOM clobbering attacks in its
-default state, but can be configured to remove `id` and `name` attributes
-through a configuration.
+default state, but can be configured to remove `id` and `name` attributes.
 
 ## XSS with Script gadgets {#script-gadgets}
 
 <em>This section is not normative.</em>
 
-Script gadgets is a technique, in which an attacker uses existing application
+Script gadgets is a technique in which an attacker uses existing application
 code from popular JavaScript libraries to cause their own code to execute.
-This is often done by injecting innocent-looking code that is only parsed and
-interpreted by a framework which then performes the execution of JavaScript
-based on that input.
+This is often done by injecting innocent-looking code or seemingly inert
+DOM nodes that is only parsed and interpreted by a framework which then
+performs the execution of JavaScript based on that input.
 
-The Sanitizer API does not try to prevent all of these attacks, but requires
-page authors to explicitly allow attributes and elements that are unknown to
-HTML and markup that is known to be widely used for templating and
-framework-specific code, like `data-` and `slot` attributes and elements like
-`<slot>` and `<template>`.
+The Sanitizer API can not prevent these attacks, but requires page authors to
+explicitly allow attributes and elements that are unknown to HTML and markup
+that is known to be widely used for templating and framework-specific code,
+like `data-` and `slot` attributes and elements like `<slot>` and `<template>`.
 We believe that these restrictions are not exhaustive and encourage page
 authors to examine their third party libraries for this behavior.
 
--- a/index.bs
+++ b/index.bs
@@ -390,6 +390,74 @@ by a custom configuration.
 {}
    ```
 
+# Security Considerations {#security-considerations}
+
+The Sanitizer API is intended to prevent DOM-Based Cross-Site Scripting
+by traversing a DocumentFragment and removing elements and attributes
+according to a configuration. The specified API must not support
+the construction of a Sanitizer object that leaves script-capable markup in
+and doing so would be a bug in the threat model.
+
+That being said, there are security issues which the correct usage of the
+Sanitizer API will not be able to protect against and the scenarios will be
+laid out in the following sections.
+
+## Server-Side Reflected and Stored XSS {#server-side-xss}
+
+<em>This section is not normative.</em>
+
+The Sanitizer API operates solely in the DOM and adds a capability to traverse
+and filter an existing DocumentFragment. The Sanitizer does not address
+server-side reflected or stored XSS.
+
+## DOM clobbering {#dom-clobbering}
+
+<em>This section is not normative.</em>
+
+DOM clobbering describes an attack in which malicious HTML confuses an
+application by naming elements through `id` or `name` attributes such that
+properties like `children` of an HTML element in the DOM are overshadowed by
+the malicious content.
+
+The Sanitizer API does not protect DOM clobbering attacks in its
+default state, but can be configured to remove `id` and `name` attributes
+through a configuration.
+
+## XSS with Script gadgets {#script-gadgets}
+
+<em>This section is not normative.</em>
+
+Script gadgets is a technique, in which an attacker uses existing application
+code from popular JavaScript libraries to cause their own code to execute.
+This is often done by injecting innocent-looking code that is only parsed and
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
...
+<em>This section is not normative.</em>
+
+DOM clobbering describes an attack in which malicious HTML confuses an
+application by naming elements through `id` or `name` attributes such that
+properties like `children` of an HTML element in the DOM are overshadowed by
+the malicious content.
+
+The Sanitizer API does not protect DOM clobbering attacks in its
+default state, but can be configured to remove `id` and `name` attributes
+through a configuration.
```
otherdaniel, line 424: editorial: Maybe strike "through a configuration". I guess the reader can figure out that configuring things happens through a configuration object.

index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
...
+
+## XSS with Script gadgets {#script-gadgets}
+
+<em>This section is not normative.</em>
+
+Script gadgets is a technique, in which an attacker uses existing application
```
otherdaniel, line 430: editorial: drop comma.

index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
...
+the construction of a Sanitizer object that leaves script-capable markup in
+and doing so would be a bug in the threat model.
+
+That being said, there are security issues which the correct usage of the
+Sanitizer API will not be able to protect against and the scenarios will be
+laid out in the following sections.
+
+## Server-Side Reflected and Stored XSS {#server-side-xss}
+
+<em>This section is not normative.</em>
```
otherdaniel, line 407: editorial: ...  not sure it makes sense to add this to every sub-section. Can we have this once or leave it out?

(I don't know what the actual rules for normatize vs non-normative are, unfortunately. :-) )

index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
...
 
+# Security Considerations {#security-considerations}
+
+The Sanitizer API is intended to prevent DOM-Based Cross-Site Scripting
+by traversing a DocumentFragment and removing elements and attributes
+according to a configuration. The specified API must not support
+the construction of a Sanitizer object that leaves script-capable markup in
+and doing so would be a bug in the threat model.
+
+That being said, there are security issues which the correct usage of the
```
otherdaniel, line 401: I wonder if we can re-phrase that... IMHO, the distinguishing factor of whether we can protect users is whether it's a browser property (i.e., the browser knows exactly which elements/attributes contain script), or whether it's an application property (e.g. whether the app will rely on `id=` attributes in a certain fashion). In this first case, the Sanitizer can and will protect. In the second case, we'll gladly supply appropriate tooling, but we cannot know whether or how to apply it.

index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
...
+Sanitizer API will not be able to protect against and the scenarios will be
+laid out in the following sections.
+
```
otherdaniel, line 404: Editorial: This is admittedly super totally redundant with your introductory paragraph, but given that pretty much everyone asks about it I'd still add an non-overridable defaults as a separate item.


Maybe something like:

## Safe by Default

The Sanitizer uses a safe-by-default allow list. Additionally, the Sanitizer has non-overridable baseline behaviour for elements and attributes the browser knows to contain script. For example, it's not possible to configure a Sanitizer to allow a `<script>` element to pass through.

index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
...
+<em>This section is not normative.</em>
+
+Script gadgets is a technique, in which an attacker uses existing application
+code from popular JavaScript libraries to cause their own code to execute.
+This is often done by injecting innocent-looking code that is only parsed and
+interpreted by a framework which then performes the execution of JavaScript
+based on that input.
+
+The Sanitizer API does not try to prevent all of these attacks, but requires
```
otherdaniel, line 436: Maybe: "does not try to prevent" => "cannot prevent these attacks in its default configuration, since it cannot not know details about the application the uses it". This is redundant with some things we said earlier, but IMHO explains it a little better.

otherdaniel (approved):
Thanks, lgtm!

Left some mostly editorial comments; feel free to adopt or ignore. :)

index.bs
mozfreddyb: duh. thanks :)

index.bs
mozfreddyb: I've seen specs where the whole _Security Considerations_ block is non-normative and some where all is normative.
I didn't want to be super normative & formal for enumerating some of the attacks, but I also did not want to weak the strong statemement at the beginning with a "non-normative" pretext.

Maybe I could group all the enumerated attacks into a non-normative section like - security considerations
-- general threat model (normative)
-- list of attacks we cant prevent (non-normative).


WDYT?

index.bs
mozfreddyb: I agree with what you say, but I don't think the application property / browser-property suggestion applies cleanly to this specific sentence. Especially given the following enumeration contains things like Server-Side XSS. Can you elaborate?

index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
...
+DOM clobbering describes an attack in which malicious HTML confuses an
+application by naming elements through `id` or `name` attributes such that
+properties like `children` of an HTML element in the DOM are overshadowed by
+the malicious content.
+
+The Sanitizer API does not protect DOM clobbering attacks in its
+default state, but can be configured to remove `id` and `name` attributes
+through a configuration.
+
+## XSS with Script gadgets {#script-gadgets}
```
koto, line 426: If you wish to link to something, we usually use https://github.com/google/security-research-pocs/tree/master/script-gadgets when describing script gadgets. https://research.google/pubs/pub46450/ would work too.

index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
...
+
+<em>This section is not normative.</em>
+
+Script gadgets is a technique, in which an attacker uses existing application
+code from popular JavaScript libraries to cause their own code to execute.
+This is often done by injecting innocent-looking code that is only parsed and
```
koto, line 432: in this context, innocent-looking DOM nodes, or seemingly inert DOM nodes.

index.bs
mozfreddyb: personally don't think it flows very well with the following sentences, but please make a different suggestion. Not opposed to the idea :)

index.bs
mozfreddyb: thanks!

index.bs
```diff
@@ -390,6 +390,74 @@ by a custom configuration.
 {}
    ```
 
+# Security Considerations {#security-considerations}
+
+The Sanitizer API is intended to prevent DOM-Based Cross-Site Scripting
+by traversing a DocumentFragment and removing elements and attributes
```
koto, line 396: Nit: Perhaps focus on a conversion of a _string_ to a DocumentFragment, instead of traversing an existing DocumentFragment. The first one is what may prevent DOM XSS (as the input string is user controlled). The creation of an inert DocumentFragment and traversing it to remove nodes is an implementation detail.

mozfreddyb:
I'm sneaking in a spell check, which is hopefully uncontroversial.

index.bs
mozfreddyb: Good point. Thanks!

index.bs
otherdaniel: My thinking is: There are clear, browser-defined script-y semantics. Those are the non-overridable defaults. Then there is everyhing else, which is not clear, browser-defined script-y semantics, like server-side XSS, or mXSS, or script gadgets, etc. In my mind, that's the dividing line between overridable and non-overridable.

I'll gladly let you decide on this, though.

index.bs
otherdaniel: Honestly, I don't know.

The whole normative vs non-normative distinction doesn't seem like a good fit for a security considerations section. The only reason I said something is that I found the repeated "non-normative" sub-headers to be a bit ugly, but that is admittedly not a valid reason. I guess I'd do whatever other specs do; except I'm not sure what that is. :)

index.bs
mozfreddyb: Thanks!

PR #75, opened 2021-04-09
Editorial changes: Use algorithm= annotations, and regularize "algorithm" language.

otherdaniel:
This implements spec feedback, that algorithms should be more explicit with the variables they use, and should use bikeshed's algorithm= annotations, which improves index generation and display.

This should be a purely editorial change and a semantic no-op.


mozfreddyb (approved):
quite the read for a no-op :sweat_smile:

Issue #69, opened 2021-03-09
Consider referencing element classes in the HTML specification


jonathanKingston:
As part of the spec, large numbers of elements are referenced in the specification that are generated with a script.

I might make sense even in a non normative sense to reference which elements / attributes are excluded / included from the default / baseline.

If this can be done at least partially by referencing [element kinds](https://html.spec.whatwg.org/#kinds-of-content) and deprecated features in the HTML specification this would make reading the specification intent easier.

otherdaniel:
Partially resolved by #78.

annevk:
This will be solved as part of merging it into the HTML Standard, no?

mozfreddyb:
Agreed. I think we can close this. It's unlike we'll merge this without, isn't it?

annevk:
Yeah, fair.

PR #78, opened 2021-04-14
Justify the built-ins.

otherdaniel:
This picks up comments from several sources, including our "spec mentor", the [TAG](https://github.com/w3ctag/design-reviews/issues/619#issuecomment-809700791), and issue #69. It provides an explanation where our defaults are coming from, and links those to the HTML spec, so that a reader can go back and forth to understand the issue.

Linking is, honestly, best effort since some elements/attributes don't have good link targets. Particularly the legacy / non-conforming ones don't seem to have a "nice" linkable definition in HTML.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -449,7 +449,7 @@ The <dfn>element kind</dfn> of an |element| is one of `regular`, `unknown`, or `
 <div algorithm="attribute kind">
 Similarly, the <dfn>attribute kind</dfn> of an |attribute| is one of `regular`
 or `unknown`. Let <var ignore>attribute kind</var> be:
-  - `unknown`, if the [[HTML]] specifcation does not assign any meaning to
+  - `unknown`, if the [[HTML]] specification does not assign any meaning to
      |attribute|'s name.
   - `regular`, otherwise.
 </div>
@@ -609,7 +609,7 @@ These constants define core behaviour of the Sanitizer algorithm.
 Note: The normative values of these constants are found below. The derivation
     of these are explained here, with an implementation in the [[DEFAULTS]]
     script. It is expected that these values will change before this
-    specifcation is finalized. Also, we expect these
+    specification is finalized. Also, we expect these
     to be updated to include additional HTML elements as they are
     introduced in user agents.
 
--- a/index.bs
+++ b/index.bs
@@ -600,6 +600,8 @@ describes, as is Internet Explorer's {{window.toStaticHTML()}}.
 
 # Appendix A: Built-in Constants # {#constants}
 
+<em>This appendix is normative, except where explicitly noted otherwise.</em>
+
 These constants define core behaviour of the Sanitizer algorithm.
 
 ## Built-ins Justification ## {#builtins-justification}
@@ -663,8 +665,6 @@ Specifically:
 
 ## The Baseline Element Allow List ## {#baseline-elements}
 
-<em>This subsection is normative.</em>
-
 The built-in <dfn>baseline element allow list</dfn> has the following value:
 
 <pre class=include-code>
@@ -674,8 +674,6 @@ highlight: js
 
 ## The Baseline Attribute Allow List ## {#baseline-attributes}
 
-<em>This subsection is normative.</em>
-
 The <dfn>baseline attribute allow list</dfn> has the following value:
 
 <pre class=include-code>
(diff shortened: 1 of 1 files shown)


iVanlIsh (approved):

index.bs
```diff
@@ -600,17 +600,71 @@ describes, as is Internet Explorer's {{window.toStaticHTML()}}.
...
-<em>This appendix is normative.</em>
-
 These constants define core behaviour of the Sanitizer algorithm.
 
-Note: The normative values of these constants are found below. They have
-    been derived with the [[DEFAULTS]] script. It is expected that these
-    values will be updated to include additional HTML elements as they are
+## Built-ins Justification ## {#builtins-justification}
+
+<em>This subsection is super duper non-normative.</em>
```
mozfreddyb, line 607: :D

index.bs
```diff
@@ -600,17 +600,71 @@ describes, as is Internet Explorer's {{window.toStaticHTML()}}.
...
+  - The {{HTMLTemplateElement}}, which introduces a new template to be used
+    by JavaScript, and its {{HTMLSlotElement}} accomplice.
+  - The frame-like [HTMLPortalElement](https://wicg.github.io/portals/).
+  - {{HTMLDataElement}},
+  - The (deprecated) [allowpaymentrequest](https://www.w3.org/TR/payment-request/) attribute.
+
+
 ## The Baseline Element Allow List ## {#baseline-element-allow-list}
 
+<em>This subsection is normative.</em>
```
mozfreddyb, line 666: do we need those hints for this section (and below)? not opposed, but it surprised me :)

mozfreddyb (approved):

index.bs
otherdaniel: No idea, honestly. My thinking was: I can no longer declare the entire section as normative/non-normative because this now differs for sub-sections. Therefore, I'll have to declare it for each subsection now.

I do wonder whether this makes sense, or wether I'm being to "algorithm-y" here. I mean, a reader could certainly figure our that if one sub-section is marked _non-normative_, the presumably the others are normative by implication.

index.bs
```diff
@@ -600,17 +600,71 @@ describes, as is Internet Explorer's {{window.toStaticHTML()}}.
...
-    been derived with the [[DEFAULTS]] script. It is expected that these
-    values will be updated to include additional HTML elements as they are
+## Built-ins Justification ## {#builtins-justification}
+
+<em>This subsection is super duper non-normative.</em>
+
+Note: The normative values of these constants are found below. The derivation
+    of these are explained here, with an implementation in the [[DEFAULTS]]
+    script. It is expected that these values will change before this
+    specifcation is finalized. Also, we expect these
```
jonathanKingston, line 612:
```suggestion
    specification is finalized. Also, we expect these
```

index.bs
```diff
@@ -600,17 +600,71 @@ describes, as is Internet Explorer's {{window.toStaticHTML()}}.
...
+  - The {{HTMLScriptElement}}, which proudly executes script as its sole purpose.
+  - All [event handler attributes]((https://html.spec.whatwg.org/#event-handler-attributes),
+    since these also execute script.
+  - {{HTMLIFrameElement}}, which loads arbitrary HTML content and therefor also script.
+  - The legacy {{HTMLObjectElement}} and {{HTMLEmbedElement}}, which load
+    non-HTML active content. Also, `<object>`'s side-kick {{HTMLParamElement}}.
+  - The [no-longer conforming](https://html.spec.whatwg.org/#non-conforming-features)
+    `<frame>`, `<frameset>`, and `<applet>` tags, which are
+    outdated versions companions of several elements listed above.
+  - The `<noframes>`, `<noembed>`, and `<nolayer>` elements.
```
jonathanKingston, line 642:
```suggestion
  - The `<noscript>`, `<noframes>`, `<noembed>`, and `<nolayer>` elements.
```

index.bs

(There was a second instance; also fixed.)

index.bs
otherdaniel: Thank you, done!

index.bs
otherdaniel: So... I went back to this, because I didn't really like this either. Now the appendix says once, "normative, except where noted otherwise", and then has the one sub-section declare itself non-normative. I think that reads well, overall...

PR #79, opened 2021-04-15
Fix reference errors.

otherdaniel:
This fixed reference errors the bnikeshed tool reports.


mozfreddyb:
Does it only complain locally? Odd that we didn't catch it in CI.

mozfreddyb (approved):

otherdaniel:
> Does it only complain locally? Odd that we didn't catch it in CI.

Don't know. It certainly _also_ complains locally.

I remember previously fixing all reference errors; but then this showed up. Maybe it's a newer version of the tool which is (more precise?

Issue #77, opened 2021-04-12
Sanitizer config introspection.


otherdaniel:
This came up during TAG review (w3ctag/design-reviews#619): _"Exposing the config on Sanitizer instances (possibly read-only) could provide an easy solution for this use case. Exposing the default config as a static property on Sanitizer might be helpful too."_ 

- A similar thought came up in #64.
- A similar mechanism existed in very early versions of our (Chrome's) implementation (`.createOptions` property)
- TAG considers whether this should be elevated to a design principle: w3ctag/design-principles#300

I think the specific question raised by TAG (how to block a specific element) can be resolved elsehow (namely by `blockElements:`), but I think the general question remains: Should we support query-ing any sanitizer instance's config, and/or the default config, with the primary use case being building a derived configuration.

iVanlIsh:
I would vote for we should since it looks like a fairly convenient feature for the developers.

mozfreddyb:
Yes! I think it would be really useful to expose the defaults (in the Sanitizer global) _and_ an instance's config (through an attribute).

otherdaniel:
Good. I've started spec-ing something; will send a PR soon-ish.

_"through an attribute"_: WebIDL spec says attributes can't be dictinonary-valued. I don't understand why; but I've picked parameter-less methods to be safe. (Am super happy to change in case I misunderstood this.)

otherdaniel:
Resolved (#80).

PR #80, opened 2021-04-19
Introspection of Sanitizer configs (and default config).

otherdaniel:
New feature: Allow introspection of Sanitizers' configuration. (#77)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -124,20 +124,31 @@ The most common use-case - preventing XSS - is handled by default,
 so that creating a Sanitizer with a custom config is necessary only to
 handle additional, application-specific use cases.
 
+Example:
+```js
+  // Replace an element's content from unsanitized input:
+  element.replaceChildren(new Sanitizer().sanitize(userControlledInput));
+```
+
+
 <pre class="idl">
   [
     Exposed=(Window),
     SecureContext
   ] interface Sanitizer {
     constructor(optional SanitizerConfig config = {});
+
     DocumentFragment sanitize(SanitizerInput input);
     DOMString sanitizeToString(SanitizerInput input);
+
+    SanitizerConfig config();
+    static SanitizerConfig defaultConfig();
   };
 </pre>
 
 * The <dfn constructor for=Sanitizer lt="Sanitizer(config)">
     <code>new Sanitizer(<var>config</var>)</code></dfn> constructor steps
-    are to create a new Sanitizer instance, and to retains a copy of |config|
+    are to create a new Sanitizer instance, and to retain a copy of |config|
     as its [=configuration object=].
 * The <dfn method for=Sanitizer><code>sanitize(<var>input</var>)</code></dfn>
     method steps are to return the result of running the [=sanitize=]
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -130,21 +130,34 @@ handle additional, application-specific use cases.
     SecureContext
   ] interface Sanitizer {
     constructor(optional SanitizerConfig config = {});
+
     DocumentFragment sanitize(SanitizerInput input);
     DOMString sanitizeToString(SanitizerInput input);
+
+    SanitizerConfig config();
+    static SanitizerConfig defaultConfig();
```
mozfreddyb, line 138: I'm surprised this is a method and not a property. Why?

index.bs
```diff
@@ -240,6 +253,32 @@ Examples:
...
+[=query the sanitizer config=] method.
+
+Examples:
+```js
+  // Does the default config allow script elements?
+  Sanitizer.defaultConfig().allowElements.includes("script")  // false
+
+  // We found a Sanitizer instance. Does it have an allow-list configured?
+  const a_sanitizer = ...;
+  !!a_sanitizer.config().allowElements // true, iff an allowElements list is configured
```
mozfreddyb, line 266: Nit: I'd spell "iff" out (or just use "if") :)

index.bs
otherdaniel: https://heycam.github.io/webidl/#idl-dictionaries says: "Dictionaries must not be used as the type of an attribute or constant," and SanitizerConfig is a dictionary.

I originally had it as an attribute, too, and then changed it when I stumbled on that sentence. If we can keep it as an attribute, I'd be happy to.  I don't understand _why_ WebIDL says this, unfortunately.

mozfreddyb: I was hoping to find an answer for the idl question by looking at their revision history, but https://github.com/heycam/webidl/commit/8d4e7ca67bcb4de476aab0411531dc97f6d17e0d didn't help me.

Either way, let's keep it as a method :)

index.bs
```diff
@@ -130,21 +130,34 @@ handle additional, application-specific use cases.
...
     as its [=configuration object=].
 * The <dfn method for=Sanitizer><code>sanitize(<var>input</var>)</code></dfn>
     method steps are to return the result of running the [=sanitize=]
     algorithm on |input|,
 * The <dfn method for=Sanitizer><code>sanitizeToString(<var>input</var>)</code></dfn>
     method steps are to return the result of running [=sanitizeToString=]
     algorithm on |input|.
+* The <dfn method for=Sanitizer><code>config()</code></dfn> method steps are
+    to return the result of running the [=query the sanitizer config=]
+    algorithm. It essentially returns a copy of the Sanitizer's
```
jonathanKingston, line 154: Should this sentence be marked as non normative?

iVanlIsh (approved):

index.bs
otherdaniel: Done.

index.bs
otherdaniel: Yeah. I solved this by removing the sentence, and instead adding an example which compares the two. That way, normativity is clear, and we gain a free example. :-)

index.bs
LeaVerou: We should not degrade API usability due to Web IDL issues. Web IDL exists to serve the Web Platform, not the other way around. [I just opened an issue](https://github.com/heycam/webidl/issues/981) in the Web IDL repo to ask about this.

LeaVerou: You could use a getter.

> It might well be that sanitizer would be better off exposing these as getters though, that's how we often expose such settings (e.g., on `ShadowRoot`).

PR #81, opened 2021-04-23
Update Readme. Move old readme to explainer.md.

otherdaniel:
The current README.md is largely written to address a standards & browser developer audience, and answers the question of why we should standardize such an API. I've moved this to explainer.md, and instead focused the README.md on the current status, and on examples of how web developers - rather than browser developers - would use the API.


otherdaniel:
Hi all... I've been a bit unhappy with the current start page, and here's a proposal how else we could present this API. Any comments? Is this going in the right direction?

mozfreddyb (approved):

Issue #71, opened 2021-03-19
Preset for navigation-less


otherdaniel:
Similar to #57, provide a preset for "navigation-less" HTML. (I.e., block `<a>` & all of its friends).

Issue #57, opened 2021-02-04
Provide presets for both "script-free" and "fetch free"


Jamesernator:
Currently the design of the sanitizer API seems to only be concerned with XSS, this means blocking all forms of scripting (e.g. `onevent`, `<script>`, etc).

However a weaker class of attacks still exist that may be used to exfiltrate data about users, and that is the ability to trigger any form of request to a third party (or sometimes even first party) resource.

For example, this image tag could be used to exfiltrate how many people view a particular piece of user generated content:

```html
<img src="https://evil.bad/honeypot.png">
```

Many other examples exist however, such as `<a ping="...">`, `<link rel="stylesheet">`, etc etc

As such I'd like to suggest having a second preset that will strip all elements/attributes that could trigger a fetch.

mozfreddyb:
Thanks for the suggestion, @Jamesernator! We'll look at offering presets for common use-cases once we get there. It's definitely in scope.

benbucksch:
+1

The sanitizer in Mozilla has this as explicitly stated goal, that loading a network resource is a bug. Network resources allow tracking by third parties, which is usually unwanted. This is true not only for email clients, but in many contexts where you want to sanitize the HTML.

Video decoders particularly are also very vulnerable to memory safety bugs, due to their complexity and performance-oriented implementation.

PR #82, opened 2021-04-23
Presets

otherdaniel:
This is more a request for comments, than an actual pull request. This isn't ready to actually go in. In fact, I'm not even sure this should be in v1 of the API.

---------------

This addresses issues #57 and #71, and specs a way for different "presets" (besides the default), by making suitable `Sanitizer` instances available as static members. Since we now have the `.config()` method, it's super easy to access & inspect their configuration, which is why I thought just offering Sanitizers ready-to-go might be a nice idea.

This CL punts on the much harder questions, namely what those preset configs should actually be like. It only looks at the mechanics of how to support those presets, once we've agreed on what they might be.


index.bs
```diff
@@ -134,6 +134,11 @@ handle additional, application-specific use cases.
     DocumentFragment sanitize(SanitizerInput input);
     DOMString sanitizeToString(SanitizerInput input);
 
+    static readonly attribute Sanitizer default;
```
mozfreddyb, line 137: I would prefer we expose this as configuration objects. That would expose some constant-defined JS objects from the Sanitizer object, e.g. Sanitizer.CONFIG_DEFAULT would return a dictionary.

That way, we wouldn't have to initialize and keep various Sanitizer isntances in every window.

index.bs
```diff
@@ -134,6 +134,11 @@ handle additional, application-specific use cases.
...
+    static readonly attribute Sanitizer nofetch;
+    static readonly attribute Sanitizer nonavigate;
```
mozfreddyb, line 139: If I may, I'll punt on the discussion of the nofetch & nonavigate use case for now, maybe even for all of v1. I fully agree that we will have to consider this at some point though :)

index.bs
```diff
@@ -168,6 +173,34 @@ Example:
...
+* <dfn attribute for=Sanitizer>default</dfn>: [=default configuration=] object.
+* <dfn attribute for=Sanitizer>nofetch</dfn>: [=nofetch configuration object=].
+* <dfn attribute for=Sanitizer>nonavigate</dfn>: [=nonavigate configuration object=].
+* <dfn attribute for=Sanitizer>richtext</dfn>: [=richtext configuration object=].
+
+Examples:
+```js
+  // Preset "rich text" allows only formatting:
+  // <p class="blubb"><b>text</b></p>
+  Sanitizer.richtext.sanitize("<p id=bla class=blubb><a href=https://example.org><b>text</b></a>");
```
mozfreddyb, line 193: With my suggestion above someone would still have to initialize it if they need it.
e.g., `myCommonMarkSanitizer = new Sanitizer(Sanitizer.CONFIG_COMMONMARK)`

index.bs
```diff
@@ -134,6 +134,11 @@ handle additional, application-specific use cases.
     DocumentFragment sanitize(SanitizerInput input);
     DOMString sanitizeToString(SanitizerInput input);
 
+    static readonly attribute Sanitizer default;
+    static readonly attribute Sanitizer nofetch;
+    static readonly attribute Sanitizer nonavigate;
+    static readonly attribute Sanitizer richtext;
```
mozfreddyb, line 140: I think it would be great if we could adopt a list of reasonable HTML elements from "somewhere else", which would allow us to inherit a widely-used set of HTML elements and thus satisfy very common use cases.
https://commonmark.org/ comes to mind, it should be somewhat easy to extract the list of html elements from their spec and provide a Sanitizer.CONFIG_COMMONMARK.

otherdaniel:
Closing outdated PR.

PR #83, opened 2021-04-26
Use .sanitize in examples by default.

otherdaniel:
Change all examples to .sanitize, except the one that motivates .sanitizeToString.

This is basically an "educational" change in that we'd prefer the use of .sanitize over .sanitizeToString, and don't want to water down that message by using .sanitizeToString as the default example throughout our spec.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -166,7 +166,22 @@ Example:
   // (For illustration purposes only. There are better ways of implementing
   // object equality in JavaScript.)
   JSON.stringify(Sanitizer.defaultConfig()) == JSON.stringify(new Sanitizer().config());  // true
+
+  // The .sanitize method is the primary API, and returns a DocumentFragment.
+  // The .sanitizeToString method returns a DocumentFragment serialized as a
+  // string.
+  (new Sanitizer().sanitize("hello")) instanceof DocumentFragment;  // true
+  typeof new Sanitizer().sanitize("hello");  // "object"
+  typeof new Sanitizer().sanitizeToString("hello");  // "string"
 ```
+
+Note: Sanitizing a string will use the [=HTML Parser=] to parse the input,
+    which will perform some degree of normalization. So even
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -166,8 +166,23 @@ Example:
   // (For illustration purposes only. There are better ways of implementing
   // object equality in JavaScript.)
   JSON.stringify(Sanitizer.defaultConfig()) == JSON.stringify(new Sanitizer().config());  // true
+
+  // The .sanitize method is the primary API, and returns a DocumentFragment.
+  // The .sanitizeToString method returns a DocumentFragment serialized as a
+  // string.
+  (new Sanitizer().sanitize("hello")) instanceof DocumentFragment;  // true
+  typeof new Sanitizer().sanitize("hello");  // "object"
```
mozfreddyb, line 174: These examples show the API, but not how they are supposed to be used
Maybe something more hands-on? I could also envision pseudo-serializing the DocumentFragment in your code comment like so
> let clean = new Sanitizer().sanitize(userdata); //  DocumentFragment [ "hello" ]
> preview.append(clean);

index.bs
```diff
@@ -179,15 +194,6 @@ parsing, cloning, or using the fragment as-is, respectively.
   typedef (DOMString or DocumentFragment or Document) SanitizerInput;
 </pre>
 
-Note: Sanitizing a string will use the [=HTML Parser=] to parse the input,
```
mozfreddyb, line 182: I would like to keep this whole block.

mozfreddyb (approved):
Looks good. Just a few nits

index.bs
mozfreddyb: How could I overlook this? :grimacing: Thanks!

index.bs
otherdaniel: Yeah. I've improved the examples a little bit.

PR #88, opened 2021-04-29
Remove textarea and title from default configuration.

otherdaniel:
Remove <textarea> and <title> from the defaults, since these have "special" parsing properties in HTML.

Based on feedback at https://bugs.chromium.org/p/chromium/issues/detail?id=1202970.


mozfreddyb:
I don't fully buy that the special parsing of textarea makes it inherently undesirable. In Firefox, we would be able to sanitize it nonetheless. But I agree that it might not be extremely useful :)

mozfreddyb (approved):

otherdaniel:
> I don't fully buy that the special parsing of textarea makes it inherently undesirable.

Yeah, I'm not entirely sure either. But the argument that the default config shouldn't make it easy to have mXSS - where the special parsing rules are a source of trouble - makes sense to me.

Note that `<textarea>` is removed from the defaults only - so you can still configure it in. This is unlike #87 `<noscript>`, which is also removed from the baseline, and hence not configurable. My thinking is that we can generously block stuff in the default, but should be more careful with the baseline.

mozfreddyb:
> > I don't fully buy that the special parsing of textarea makes it inherently undesirable.
> 
> Yeah, I'm not entirely sure either. But the argument that the default config shouldn't make it easy to have mXSS - where the special parsing rules are a source of trouble - makes sense to me.
>

I suppose we should rethink the whole topic of mXSS then, because we don't fully acknowledge it as part of our thread model per the Security Considerations section.

> 
> Note that `<textarea>` is removed from the defaults only - so you can still configure it in. This is unlike #87 `<noscript>`, which is also removed from the baseline, and hence not configurable. My thinking is that we can generously block stuff in the default, but should be more careful with the baseline.

Fully agreed, hence the review+.

PR #90, opened 2021-05-12
explicit conversion of element names to lowercase

mozfreddyb:
this came up during implementation/review and I think it would be useful to make explicit. Happy to incorporate it somewhere else or in a different form.

This is the best place I could find, but maybe you'll find a better section of the document :-)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -225,6 +225,9 @@ is used as the configuration object.
   };
 </pre>
 
+Note: Element names are expected to be ascii lowercase and those that don't
+  conform will be lowercased.
+
 : allowElements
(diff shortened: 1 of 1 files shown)


otherdaniel (approved):
1. Content: Looks good.
    (On our end, I had implemented it so that we'd return whatever the user had given us, and would force the case only internally. Either way works, though.)

2. Formalization: Since we're doing all this claptrap with algorithms-language, this should really be reflected in the algorithm descriptions. That said, I'll be happy to re-word it in algorithm-y fashion whenever I get to do the next editing round; this shouldn't block us now.

PR #95, opened 2021-05-26
Update .gitignore.

otherdaniel:
Ignore .ninja files and out/ directory.


otherdaniel:
This isn't necessary, but makes my live slightly easier.

PR #96, opened 2021-05-26
Remove deploy_key file.

otherdaniel:


otherdaniel:
Honestly, no idea whether this is correct, but I couldn't find any reference to the deploy_key.enc file left. It was used by the older, deploy.sh script.

PR #97, opened 2021-05-27
Normalize element names.

otherdaniel:
Re-word and reflect the changes from PR #90 in the algorithms and examples.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -292,7 +292,7 @@ handle additional, application-specific use cases.
 
 * The <dfn constructor for=Sanitizer lt="Sanitizer(config)">
     <code>new Sanitizer(<var>config</var>)</code></dfn> constructor steps
-    are to run the steps [=create a sanitizer=] algorithm with |config| as
+    are to run the [=create a sanitizer=] algorithm steps with |config| as
     parameter.
 * The <dfn method for=Sanitizer><code>sanitize(<var>input</var>)</code></dfn>
     method steps are to return the result of running the [=sanitize=]
@@ -497,6 +497,10 @@ Note: `allowElements` creates a sanitizer that defaults to dropping elements,
   elements. Using both types is possible, but is probably of little practical
   use. The same applies to `allowAttributes` and `dropAttributes`.
 
+Note: Element names are normalized, following the rules of the HTML Parser.
+  This means elements are usually lowercase, except for a small-ish number
+  of mixed case element names in non-HTML namespaces (SVG, MathML).
+
 <div class="example">
 ```js
   const sample = "Some text <b><i>with</i></b> <blink>tags</blink>.";
@@ -571,8 +575,8 @@ A sanitizer's configuration can be queried using the
 An <dfn>attribute match list</dfn> is a map of attribute names to element names,
 where the special name "*" stands for all elements. A given |attribute|
 belonging to an |element| matches an [=attribute match list=], if the
-attribute's local name is a key in the match list, and element's local name
-or `"*"` are found in the attribute's value list.
+attribute's [=Attr/local name=] is a key in the match list, and element's
+[=Element/local name=] or `"*"` are found in the attribute's value list.
 
 <pre class="idl">
   typedef record&lt;DOMString, sequence&lt;DOMString>> AttributeMatchList;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -461,9 +462,14 @@ run these steps:
   3. Let |result| be a newly constructed SanitizerOptions dictionary.
   4. For any non-empty member of |config| whose key is declared in
      SanitizerOptions, copy the value to |result|.
-  5. Return |result|.
+  5. All element names in |result| should be converted to lower case.
```
mozfreddyb, line 465: I don't think we should convert when getting the config, but rather when accepting the config during construction.

otherdaniel: Hmm. I don't think this makes an observable difference. But changes anyways...

Related: I was trying to figure out whether this would save me from specifiying comparisons as case-independent. What I find in DOM is that [tagName is always upper-cased](https://dom.spec.whatwg.org/#dom-element-tagname) and is arguably the wrong thing anyhow because we should compare to localName, and localName doesn't seem to have any case guarantees. With that, I do wonder why we even want force-lowercase in the first place.

mozfreddyb: hm, this is funny. I thought I had a reference where localname is always lower-cased. But now I can't find it anymore... 
@hsivonen brought this up in code review. Maybe he remembers?

otherdaniel: Note to self: I had assumed tagName and local name were the same, which they aren't. tagName is a legacy concept and uppercase. local name is the current concept and (mostly) lowercase.

The closest definition I can find is: https://dom.spec.whatwg.org/#dom-document-createelement

This says that local names are always lower-cased in HTML documents (but not XML ones). I'm going to assume that we'll forfit XML docs and just always lowercase. I'll change the PR to consistently refer to local names, and not to tag names. (Also the spec, in case I'll find additional references there.)

Andrew-Cottrell: Might there be an issue if the sanitizer were to support [SVG elements](https://developer.mozilla.org/en-US/docs/Web/SVG/Element), many of which have mixed-case `localName` values?

For example,

```javascript
document.createElementNS("http://www.w3.org/2000/svg", "linearGradient").localName; // "linearGradient"
```

In my testing with the Chrome 91 browser, the `localName` is mixed-case even when the SVG element are embedded in an HTML document.

```javascript
document.querySelector("linearGradient").localName; // "linearGradient"
```

otherdaniel: Sorry I'm a bit late, I just wanted to confirm that this is indeed a problem for SVG & MathML elements.

Right now, we're cheating by simply ignoring non-HTML. But clearly SVG & MathML are on the agenda. So we need to make sure to not inadvertently block our options there.

hsivonen: The HTML tokenizer always lower-cases incoming element names. For SVG, it then [re-adjusts the case for a list of elements](https://html.spec.whatwg.org/#parsing-main-inforeign:svg-namespace).

Does the Sanitizer API consider allow-listing elements that are in a namespace other than HTML, SVG, and MathML to be in scope for the API?

mozfreddyb: > Does the Sanitizer API consider allow-listing elements that are in a namespace other than HTML, SVG, and MathML to be in scope for the API?

Strong "No!" :)

otherdaniel: Thanks all for the feedback. Very extra thanks fo @hsivonen for the spec reference, because I think that provides us with a good way forward:

Instead of prescribing either lower or uppercase, we'll spec it as normalize-in-whatever-fashion-HTML-does. That should match the principle of least surprise, and also allow for easy implementation, since presumably all user agents have the suitable normalization already implemented somewhere.

otherdaniel:
Updated the PR:
- Replace "lower case" with normalize.
- Add a nromalization algorithm. As we currently don't support SVG & MathML anyhow, the algorithm consists of lower-casing. I've included the proper version - with SVG + MathML support - in the 'Issue:'.
- Consistently replaced all references from 'tag names' with 'local names', including proper spec references.
- Added a note that - since we're now matching & referencing HTML spec normalization - normalization + name comparison is largely equivalent to case-insensitive comparison.

mozfreddyb:
@otherdaniel Can you please rebase this PR to get it landed?

otherdaniel:
> @otherdaniel Can you please rebase this PR to get it landed?

Done. But there is this new-ish bot that still doesn't want the PR. It looked at [their docs](https://github.com/w3c/ash-nazg/), and they suggest we require certain files (w3c.json, LICENSE.md, CONTRIBUTING.md), which we have. It seems that one needs to register the repo with them first, but off-hand I didn't find how to do so. :(

mozfreddyb:
That's super odd. I filed https://github.com/w3c/spec-prod/issues/96 and hope that's indeed the cause of our issue.

mozfreddyb:
Asking `ipr` to revalidate the pull request, it's telling me "PR not found: WICG/sanitizer-api/pulls/97", which seems...nonsense?
Can you help @deniak?

marcoscaceres:
Weird about the IPR bot. Was this repo moved or renamed recently? I’ll try a few things to see if I can get it going again tomorrow.

deniak:
> Asking `ipr` to revalidate the pull request, it's telling me "PR not found: WICG/sanitizer-api/pulls/97", which seems...nonsense?
> Can you help @deniak?

That repository is [not known to the repository manager](https://labs.w3.org/hatchery/repo-manager/repos). Someone with write access to the repo and with admin privileges to the repo manager should add it first. I've just asked @plehegar to add it.

deniak:
Ok, @plehegar confirmed the repository was renamed from `purification`. I'll try to reflect the change to the repo manager.

deniak:
The IPR check is now fixed: https://labs.w3.org/repo-manager/pr/id/WICG/sanitizer-api/97

mozfreddyb:
thanks @deniak

PR #106, opened 2021-07-01
Specify an option to allow/drop comment nodes.

otherdaniel:


mozfreddyb:
:+1: , I think the use-case makes sense and the configuration seems unproblematic. This is, realistically, something we could easily implement in Firefox.

However, the example uses the outdated `sanitize()` function which accepts a string...I suppose that needs to change?

otherdaniel:
> However, the example uses the outdated sanitize() function which accepts a string...I suppose that needs to change?

Good point, I had overlooked that. However, that change that actually outdates `sanitize(DOMString)` isn't in yet... I'll just delay this change until the new strings are in, and will then update the examples.

mozfreddyb (approved):

koto:
Don't HTML comments explicitly enable mXSS when serializing to a string as you can have a comment with `-->` inside? It's probably not a very important issue after the removal of the API outputting strings, but still might be a problem if the libraries end up serializing the sanitized nodes to strings and back (which I suspect existing sanitizers would be forced to do).

koto:
Heh, disregard my previous comment. This could only ever be an issue if the input to the sanitizer is a node (whereas so far there are only string inputs).

otherdaniel:
> Heh, disregard my previous comment. This could only ever be an issue if the input to the sanitizer is a node (whereas so far there are only string inputs).

Well, it's a valid concern, but it does require quite a few contortions to become an issue. In the new-strings-world, we either stay in nodes-only land, or we convert a string to a node. So for this to be a problem, one would have to constuct an offending node, then sanitize it (with comments explicitly enabled), then un-parse it, and then re-sanitize it with string->node functions.

PR #110, opened 2021-07-09
Explain string APIs.

otherdaniel:
Human-readable explanation of the strings API, and why they are the way they are. Including exapmles and references.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -43,6 +43,14 @@ spec:dom; type:dfn; text:append
     "title": "mXSS Attacks: Attacking well-secured Web-Applications by using innerHTML Mutations",
     "publisher": "Ruhr-Universität Bochum"
   },
+  "MXSS1": {
+    "href": "https://research.securitum.com/mutation-xss-via-mathml-mutation-dompurify-2-0-17-bypass/",
+    "title": "Mutation XSS via namespace confusion"
+  },
+  "MXSS2": {
+    "href": "https://www.checkmarx.com/blog/technical-blog/vulnerabilities-discovered-in-mozilla-bleach/",
+    "title": "CVE-2020-6802 Write-up"
+  },
   "HTML":{
     "authors": [
       "Anne van Kesteren",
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -114,6 +122,154 @@ let sanitizedFragment = s.sanitize(userControlledInput);
...
 
+## The Trouble With Strings ## {#strings}
+
+Many HTML sanitizer libraries are based on string-to-string APIs, while this
+API does not offer such a method directly. This sub-section explains the
+reasons and implications for the Sanitizer API.
+
+To convert a string into a tree of nodes (or a fragment), it needs to be parsed.
+The [HTML parsing algorithm](https://html.spec.whatwg.org/multipage/parsing.html#parsing)
+carefully specifies how parsing HTML worksThis parsing algorithm is dependent
```
mozfreddyb, line 133: Nit: `s/worksThis/works. This/`

index.bs
```diff
@@ -114,6 +122,154 @@ let sanitizedFragment = s.sanitize(userControlledInput);
...
+take the sanitization result and remove it from its context. In this case, the
+responsibility to prevent mXSS-class attacks that stem from mis-applying those
+strings in an inappropriate context remains with the developer.
+
+<div class="example">
+```js
+const user_string = "...";  // The user string.
+const sanitizer = new Sanitizer( ... );  // Our Sanitizer.
+
+// We plan to insert this string into a <div> container, but for $REASONS we
```
mozfreddyb, line 265: I think it's useful to have this use-case covered even though we discourage it.
To further emphasize that this is a discouraged use-case, let's rewrite this from a "We" to a "The developer" point-of-view, OK?

E.g.,

```
// The developer plans to insert this string into a <div> element, but has to
// keep this around as a string (instead of an element). It's important that the
// developer remembers the parsing context and MUST NOT use this in
// a different context in order to prevent mXSS attacks.
```

or something along those lines.

mozfreddyb (approved):
Great rewrite, thanks a lot!

That reads to me as if we can't rewrite the mXSS-bits in the Security Considerations section then. I'm pretty stoked that we may actually include this in our threat model and come up with a good response to mXSS.

index.bs

index.bs

mozfreddyb (approved):

PR #111, opened 2021-07-09
Link to WPT tests.

otherdaniel:
No idea how useful this is... I found this functionatliy in the bikeshed manual (https://tabatkins.github.io/bikeshed/#wpt-element), and its easy to do. I think the main drawback is that presently the test suite isn't super impressive, so I'm not sure how usefule this is.

wdyt?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -16,6 +16,8 @@ Indent: 2
 Work Status: exploring
 Boilerplate: omit conformance
 Markup Shorthands: css off, markdown on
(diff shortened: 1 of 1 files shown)


mozfreddyb:
> I think the main drawback is that presently the test suite isn't super impressive, so I'm not sure how usefule this is.

I'm positive we will overcome this step by step.

mozfreddyb (approved):

PR #120, opened 2021-08-11
Fix `blockAttributes` to `dropAttributes` in sample code and nits

Jxck:
`blockAttributes` is not in the spec, it should be a `dropAttributes`.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -505,13 +505,13 @@ Note: `allowElements` creates a sanitizer that defaults to dropping elements,
   const sample = "Some text <b><i>with</i></b> <blink>tags</blink>.";
 
   // Some text <b>with</b> text tags.
-  new Sanitizer({allowElements: [ "b" ]).sanitize(sample);
+  new Sanitizer({allowElements: [ "b" ]}).sanitize(sample);
 
   // Some text <i>with</i> <blink>tags</blink>.
-  new Sanitizer({blockElements: [ "b" ]).sanitize(sample);
+  new Sanitizer({blockElements: [ "b" ]}).sanitize(sample);
 
   // Some text <blink>tags</blink>.
-  new Sanitizer({dropElements: [ "b" ]).sanitize(sample);
+  new Sanitizer({dropElements: [ "b" ]}).sanitize(sample);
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -593,10 +593,10 @@ Examples for attributes and attribute match lists:
   new Sanitizer({allowAttributes: {"style": ["*"]}}).sanitize(sample);
 
   // Block <span id>: <span class='theclass' style='font-weight: bold'>...</span>
-  new Sanitizer({blockAttributes: {"id": ["span"]}}).sanitize(sample);
+  new Sanitizer({dropAttributes: {"id": ["span"]}}).sanitize(sample);
```
mozfreddyb, line 596: The missing braces above looked correct, but why did you cahnge block with drop here?

index.bs
```diff
@@ -593,10 +593,10 @@ Examples for attributes and attribute match lists:
...
 
   // Block id, everywhere: <span class='theclass' style='font-weight: bold'>...</span>
-  new Sanitizer({blockAttributes: {"id": ["*"]}}).sanitize(sample);
+  new Sanitizer({dropAttributes: {"id": ["*"]}}).sanitize(sample);
```
mozfreddyb, line 599: (same question as above)

mozfreddyb (changes requested):

index.bs
Jxck: because Spec doesn't has `blockAttributes` so I thought its a miss.
https://wicg.github.io/sanitizer-api/#config

index.bs
Jxck: (same answer above)

index.bs
mozfreddyb: oops...fair point :)
Can you change the comment above to also say `drop`?

Jxck: Done, is that works ?

mozfreddyb: Maybe you did not push yet? The code comment still says "block" for me.

Jxck: Oh, my fault. Now its on!

mozfreddyb (approved):

Issue #153, opened 2022-06-10
Consider creating an "effective sanitizer configuration" during construction


mozfreddyb:
I noticed that we do a lot of if/else dance during sanitization that make the algorithms a bit less clear and are sometimes brittle.

I suggest we separate the concern of "taking a config" form the concern of sanitization by building an config dictionary based on the supplied arguments (and filling the missing properties with defaults).

Then we only need to look into that one specific config during sanitization, instead of doing the "is it present, if not look elsewhere" dance.

I believe we discussed this before, but it occurred to me more closely when re-reviewing the spec.

otherdaniel:
For a while, the spec was written that way, and for the reasons you give. I dropped it, since it didn't seem to simplify anything after all: I needed all the same conditions and checks, I just had them in a different place. That way, the "effective configuration" was merely an indirection rather than a valid abstraction. So all it accomplished was that when figuring out what to do, I had to additionally look in another place.

Would be happy to try again, if we have a good take on how to define an effective configuration.

mozfreddyb:
This should be resolved due to the config canonicalization coming up in #208. This is mostly resolved (and there is agreement).

PR #157, opened 2022-06-10
Review nits and issues in-place

mozfreddyb:
This patch changes variou editorial things, from missing `|` to improved iterations.
There are also some "actual" changes, where instead of "removing an attribute", we're actually
invoking the removal algorithm with the right parameters.

I've tried to remove most consequential or controversial things into their own issues.
That being said, I have introduced lots of `Issue:` stuff direclty into the spec
imho this has the benefit of fixing things correctly and cohesively in the right place.

I'm btw volunteering to fixing most of those, if nobody else beats me to it.
So, while this is a glaring red "ISSUE" in the HTML view, it's also mostly
to be considereda "Freddy's note to self".


index.bs
```diff
@@ -792,77 +798,84 @@ run these steps:
...
     1. Return `drop`.
   1. If |node| is a {{Text}} [=node=]: Return `keep`.
-  1. Return `drop`
+  1. [=Assert=]: |node| is a {{ProcessingInstruction}}
+  1. Return `drop`.
 </div>
 
+Issue(151): The [=sanitize action for an attribute=] algorithm parameters do not match.
+Issue(152): consider renaming "configuration object" to "configuration dictionary".
+Issue(153): consider creating an effective sanitizer config. Also, IDL guarantees that a config is ALWAYS given. The question is really whether the members exists.
```
otherdaniel, line 877: What would be the consequence of this?

otherdaniel (approved):
Thanks. This is a much needed scrub of the spec. :-)

index.bs
mozfreddyb: I haven't fully fleshed it out. Let me make a PR towards that idea in #153 if you don't mind.

index.bs
```diff
@@ -727,6 +729,8 @@ To <dfn>normalize element name</dfn> |name|, run these steps:
   1. Return `null`.
 </div>
 
+  Issue(145): we shouldnt normalize lowercase/uppercase for element names at all. Other APIs are also case-sensitive (e.g., MutationObserver)
```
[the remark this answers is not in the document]
mozfreddyb: Added the full stop. Let's discuss in the issue itself. I've toned down the note into "we should find out" instead of "we shouldnt normalize", for the sake of getting this rather large collection of changes ready to land

PR #159, opened 2022-06-27
Introduce allowUnknownElements config option.

otherdaniel:
- Introduce allowUnknownElements config option, based on recent discussions in the most recent meeting.
- Some drive-by fixes to silence Bikeshed warnings & error messages.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -315,6 +315,7 @@ applies a string using a `Sanitizer` directly to an existing element node.
   dictionary SetHTMLOptions {
     Sanitizer sanitizer;
   };
+  [SecureContext]
   partial interface Element {
     undefined setHTML(DOMString input, optional SetHTMLOptions options = {});
   };
@@ -323,6 +324,12 @@ applies a string using a `Sanitizer` directly to an existing element node.
 * The <dfn method for=Element><code>setHTML(<var>input</var>, <var>options</var>)</code></dfn>
   method steps are to run the [=sanitizeAndSet=] algorithm on [=this=], |input|, and
   |options|.
+
+<wpt>
+sanitizer-secure-context.https.html
+sanitizer-insecure-context.html
+idlharness.https.window.js
+</wpt>
 
 Issue: Is this how we specify a method on existing class "owned" by a different spe?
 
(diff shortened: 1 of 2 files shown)


otherdaniel:
WPT tests upcoming in https://github.com/web-platform-tests/wpt/pull/34602.

This is elements only. Would this also apply to attributes?

mozfreddyb:
Let's include attributes and rename to `allowUnknownMarkup`.

Andrew-Cottrell:
> > Let's include attributes and rename to `allowUnknownMarkup`.
> 
> I checked and there's no other great terminology here. We either go fully verbose `allowUnknownElementsAndAttributes` or vague `allowUnknown` or inventive `allowUnknownMarkup`. I have a _slight_ tendency not to invent something new but no strong preference. 😕

Another potential option is `allowUnknownNodes`, based on the fact both [`Element`](https://dom.spec.whatwg.org/#interface-element) and [`Attr`](https://dom.spec.whatwg.org/#interface-attr) implement the [`Node`](https://dom.spec.whatwg.org/#interface-node) interface. There are other `Node` types, but I assume they either would not be relevant (e.g. document fragment) or would not be described as _unknown_ (e.g. text).

Andrew-Cottrell:
> Comments are also nodes, we already have allowComments.

That's potentially also a point against `allowUnknownMarkup` as _markup_ could be understood to includes comments.

Sora2455:
Why not have two options then, allowUnknownElements and allowUnknownAttributes?

mozfreddyb (changes requested):
Let's settle on a term that includes attributes

otherdaniel:
> > Why not have two options then, allowUnknownElements and allowUnknownAttributes?
> 
> The security ramifications are the same.
> 
> imho there's no point in just allowing one or the other. In the end, we'll still require users to explicitly list the element & attribute names in the allow lists anyway.

I agree on this bit. Elements and attributes have the same implications here, so I'd also think they should be handled by the same setting.

I don't care strongly about any of the names; we just have to pick one. .)

IMHO, mild preference for `allowUnknown` or `allowUnknownMarkup`; I think I'd have understood what those do without looking up the spec. But `allowUnknownElementsOrAttributes` or `allowUnknownNodes` would work, too. I guess one could also pick `allowUnknownNames` or `allowUnknownQNames`, since this is really about named entities in the DOM.

mozfreddyb (approved):
Thank you!

otherdaniel:
I'll land this in its current form, with `allowUnknownMarkup`. WPT tests are on their way [over here](https://github.com/web-platform-tests/wpt/pull/34602).

PR #161, opened 2022-07-04
Invoking the right 'replace' function when sanitizing a DocFragment

### LINKED ISSUE #156: replacing a document with its children needs to be specified more precisely
Currently:
```    1. If |m|[|node|] is `block`, [=/replace=] |node| with all of its element and text node children from |fragment|.```

I believe we need to to explicitly iterate over the children and insert into parent. It could collect them in a variable or do things in place, but this is a bit too imprecise.

### PR DESCRIPTION
**Pull request by mozfreddyb**
(Merge the other pull request in 152 before this one)

Invoke the right 'replace' function when sanitize a documentfragment - fixes #156

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -784,7 +784,7 @@ run these steps:
   1. Let |fragment| be the result of invoking the
      [html fragment parsing algorithm](https://w3c.github.io/DOM-Parsing/#dfn-fragment-parsing-algorithm)
      with |this| as the `context node` and |value| as `markup`.
-  1. Run the steos if the [=sanitize a document fragment=] algorithm
+  1. Run the steps if the [=sanitize a document fragment=] algorithm
      on |fragment|, using |sanitizer| as the current {{Sanitizer}} instance.
   1. [=Replace all=] with |fragment| as the `node` and |this| as the `parent`.
 <wpt>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by yoavweiss** [comment]
Hey @mozfreddyb! Can you join the WICG to appease the IPR bots? :)

**Comment by mozfreddyb** [comment]
Pretty sure, I've been in the WICG for a long while now. Revalidating fixed it. Hm.

File: index.bs [hunk 1]
```diff
@@ -842,7 +842,7 @@ To <dfn>sanitize a document fragment</dfn> named |fragment| with a {{Sanitizer}}
     1. [=map/Set=] |m|[|node|] to |action|.
   1. [=list/iterate|For each=] |node| in |nodes|:
     1. If |m|[|node|] is `drop`, [=/remove=] |node|.
-    1. If |m|[|node|] is `block`, [=/replace=] |node| with all of its element and text node children from |fragment|.
+    1. If |m|[|node|] is `block`, invoke [=Replace with=] with |node| as |this| and |node|'s [=tree/children=] as |nodes|.
```
otherdaniel, line 845: Hmm. As I read `[=/replace=]`, it explicitly handles the case of a DocumentFragment (and does the right thing for us, too). I think just dropping the "with all of its element and text node children" language and just using |fragment| as |node| should be correct.

`[=ReplaceWith=]` -- if I found the correct `replaceWith`, which I'm not sure about -- calls into `converting nodes into a node`, which creates a DocumentFragment, and then calls `replace` with it. If so, then this would just add extra steps by taking the nodes out of a fragment; then putting them into a new fragment; and then calling replace, which takes the nodes out of their new fragment again and puts them into the DOM. Technically correct, but arguably not easier to understand.

---

General question: Is there a convenient method to resolve these spec references? Best I can think of right now is patching in the change locally; running bikeshed; and then seeing where it takes me.

yoavweiss:
@mozfreddyb - apologies for the bots then! I should've checked..

index.bs
otherdaniel: Well, invoking `replace` without a parent reference would work just how `[=replaceWith=]` does it, by `let ... be |node|'s [=parent=]`.

I now notice that the use of `|fragment|` in the current text is plainly wrong, however, and this indeed needs to be the `[=tree/children=]` of `|node|`. And that requires creating a separate DocumentFragment anyway, as `[=replaceWith=]` implicitly does. So... let's go with `[=replaceWIth=]`.

otherdaniel (approved):

PR #162, opened 2022-07-07
nit: Remove bikeshed warning about single-use variable.

otherdaniel:
Bikeshed issues a warning if a variable has not been defined, and we do not define |this|.

> WARNING: The following \<var>s were only used once in the document:
>   'this'
> If these are not typos, please add an ignore='' attribute to the \<var>.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -292,7 +292,7 @@ handle additional, application-specific use cases.
 
 * The <dfn constructor for=Sanitizer lt="Sanitizer(config)">
     <code>new Sanitizer(<var>config</var>)</code></dfn> constructor steps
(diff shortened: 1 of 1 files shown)


mozfreddyb:
:+1: for now. We _do_ need to define and make `|this|` work eventually.

otherdaniel:
Hmm. I looked at some other standards, and they don't markup `this` as a variable, but as a reference to https://webidl.spec.whatwg.org/#this. Maybe that'd work?

mozfreddyb:
Yeah, probably that's it. In the end, from what I understand what @annevk said was that a constructor operates on a `this` and the construction is somewhat implicit

PR #164, opened 2022-07-11
Make `this` a spec reference.

otherdaniel:
Follow-up to #162: Mark up `this` as a reference, rather than a locally defined variable.


otherdaniel:
Hmm. On second thought: Maybe we should just drop the "this" parameter, and instead replace the reference downstream?

mozfreddyb:
FWIW I took this as a reference https://url.spec.whatwg.org/#constructors

mozfreddyb:
i.e., the *sanitizer* mentioned in https://wicg.github.io/sanitizer-api/#create-a-sanitizer needs to be that _this_

PR #165, opened 2022-07-11
Explainer update to use .setHTML method.

otherdaniel:


README.md
```diff
@@ -42,10 +42,12 @@ If you wish to try out early Sanitizer implementations, the
...
-The core API of the Sanitizer is rather simple: There are `Sanitizer` objects,
-and they have a `.sanitize` method that produces a `DocumentFragment` from a
-given text string. Or from another `DocumentFragment`, or even from a whole
-`Document`.
+The core API of the Sanitizer is rather simple: There are `Sanitizer` objects
+which encapsulate the information of what the sanitizer is supposed to do,
+and there is an `Element.setHTML` method which takes a string and (optionally)
+an options dictionary with a `"sanitizer"` key and a `Sanitizer` instance as
+a value. `.setHTML` will parse and sanitize the string it receives, and will
+update the children of the Element it is called on with the result.
```
mozfreddyb, line 50: Quite the long sentences. Can we cut the complexity by making it grow gradually?
How about something along those lines:
* This API prevents XSS, which means sanitizing a tree of DOM nodes
* There's a safe default that states which should be removed and kept
* if all you care about is XSS, you may just use `someElement.setHTML(untrusted_input)`
* if you need to customize, use a Sanitizer constructor to hold the configuration and pass it in

Would that make sense?

README.md
```diff
@@ -42,10 +42,12 @@ If you wish to try out early Sanitizer implementations, the
...
 
 Example:
 ```js
```
mozfreddyb, line 53: While we're at it, I think we could start explaining/defining two variables for an element and for untrusted input a priori. Repeating the `document.querySelector()` call so often is a bit noisy, isn't it?

README.md
```diff
@@ -92,14 +99,19 @@ Example:
   });
 
   const untrusted_example = "Well, <em class=nonchalant onclick='alert(\'General Kenobi\');'><a href='https://obiwan.org/home.php'>hello there<a>!"
```
mozfreddyb, line 101: obiwan.org seems to be an existing site? Can we use obiwan.example?

README.md
otherdaniel: Done.

README.md
otherdaniel: Done.

(I can't find a page or whois entry to obiwan.org, but let's use the reserved .example to be safe.)

mozfreddyb (approved):
thank you!

PR #166, opened 2022-07-17
Fixed the error in the description of the allowElements and the dropElements

najmiehsa:
There are two errors in the provided description in the note section here https://wicg.github.io/sanitizer-api/#attribute-drop-list
this is a fix for those errors.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -515,8 +515,8 @@ is used as the configuration dictionary.
 :: The <dfn>allow comments option</dfn> determines whether HTML comments are
    allowed.
 
-Note: `allowElements` creates a sanitizer that defaults to dropping elements,
-  while `blockElements` and `dropElements` defaults to keeping unknown
+Note: `allowElements` creates a sanitizer that defaults to keeping elements,
+  while `blockElements` and `dropElements` defaults to dropping unknown
   elements. Using both types is possible, but is probably of little practical
   use. The same applies to `allowAttributes` and `dropAttributes`.
 


benbucksch (changes requested):
Sorry, but I don't understand this change. The old wording makes sense to me. The new one does not.

If I understand `allowElements` correctly, it allows only those element specifically listed. E.g. if `allowElements` contained only `html body p`, then a `link` would not be allowed. Therefore, the old wording of "defaults to dropping" (i.e. "link" is removed from the sanitized document) is correct and the proposed wording "defaults to keeping" (i.e. "link" is in the sanitized document) is incorrect.

mozfreddyb:
I believe you are right @benbucksch, but I also believe the current wording is quite ambiguous - at least without context. :)

otherdaniel:
When writing that, I certainly _intended_ the meaning Ben reads out of it: If you specify an `allowList`, then the default (for the other elements, not in the list) is "`drop`". But if that isn't what everyone reads out of it, then we should re-word for clarity.

I think that additionally, with #159, the underlying situation has now changed a bit, in that we now default-`drop` much more than we used to.

I think the comment that an `.allowList` and a `.dropList` at the same doesn't make much sense still holds, while the other bits aren't that interesting. How about rewriting this to specifically focus on combined allow and drop/block lists, and maybe adding an example?

hamishwillee:
@otherdaniel I've been redirected here based on a similar discussion on MDN: https://github.com/mdn/content/pull/18264#issuecomment-1193488751

I had no problem understanding the current wording. Normally I'd argue as you have that if the content can be misinterpreted by anyone it needs more work, but I really can't see how this is in any way ambiguous. 
If you had to, maybe you could expand the case a bit

> Note: `allowElements` creates a sanitizer that will drop any elements that are not in the `allowElements` list, 
  while `blockElements` and `dropElements` create a sanitizer that will allow all elements except those that are in the  properties.
  
 However I do have a problem with the last part:
 
 > Using both types is possible, but is probably of little practical use. The same applies to `allowAttributes` and `dropAttributes`.
 
We have already defined that `allowElements`  will drop any elements that are not in the property and  `blockElements` and `dropElements`  will allow any elements but remove those in the property.

So what if the same element is in `allowElements`  and `dropElements` ?

What if we have just one element in  `allowElements` "a" and  `dropElements` has "p" (randomly chosen). Does the fact that allow elements drops all elements other than a mean dropElements entry is pointless? I guess you can see the kind of questions in my mind.

otherdaniel:
> So what if the same element is in `allowElements` and `dropElements` ?

If you have `allowElements: ["a", "b"]` and `dropElements: ["b", "c"]`, and your input contains elements `a`, `b`, `c`, and `d`, then only `a` would be allowed:  `b` and `c` get dropped when the drop list is processed, and `d` gets dropped when the allow list is processed. `c` would get dropped when the allow list is processed, too, but the spec handles the drop list first and thus `c` won't get this far. [[Ref]](https://wicg.github.io/sanitizer-api/#sanitize-action-for-an-element).

If one views these as sets, then an allow list `A` with a drop list `B` is equivalent to an allow list `A without B` (`A \ B`) and no drop list.

I think this _might_ be confusing to developers, which is why I wanted to add the note. Whenever you have an allow _and_ a drop list, there is a smaller and easier to read equivalent configuration one could have written instead.

najmiehsa:
Based on this feedback, should I add examples here?

hamishwillee:
Thanks @otherdaniel, great explanation. What wasn't clear to me was that drops are processed before allows. What you have done might be enough for a specification, where the readership is expected to pull that kind of fact from elsewhere.

For MDN I will do something like:

> **Note:** `allowElements` creates a sanitizer that will drop any elements that are not in `allowElements`,
while `blockElements` and `dropElements` create a sanitizer that will allow all elements except those in these properties.
>
> `blockElements` and `dropElements` are processed before `allowElements`.
> If you specify both properties, the elements in `blockElements` or `dropElements` will be discarded first, followed by any elements not in `allowElements`.
> So while it is possible to specify both types of properties at the same time, the intent can always be more clearly captured using just one type.
>
> The same applies to `allowAttributes` and `dropAttributes`.

PR #174, opened 2022-09-09
Fix spec references: this and template contents.

otherdaniel:
Fix up two spec references, to silence bikeshed errors:
- |this| => [=this=]
- define a reference for "template contents"


otherdaniel:
This replaces the very similar #164, which I closed. I'd just like to get this in to make bikeshed happy. We need a separate scrub of the spec text to see whether we should define some of these things differently.

PR #176, opened 2022-09-09
Properly define "block" action.

otherdaniel:
The "[=replace with=]" spec ref doesn't actually exist. Replace this with a combination of append and replace.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -851,7 +851,9 @@ To <dfn>sanitize a document fragment</dfn> named |fragment| with a {{Sanitizer}}
     1. [=map/Set=] |m|[|node|] to |action|.
   1. [=list/iterate|For each=] |node| in |nodes|:
     1. If |m|[|node|] is `drop`, [=/remove=] |node|.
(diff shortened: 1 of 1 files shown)


otherdaniel:
Together with #174 + #175, this PR should allow warning-free assembly of the spec again.

mozfreddyb (commented):
I think the intention was to invoke an existing function. Could it be https://dom.spec.whatwg.org/#concept-node-replace-all?

otherdaniel:
> I think the intention was to invoke an existing function. Could it be https://dom.spec.whatwg.org/#concept-node-replace-all?

Yes, but on a close look I don't think any of the existing algorithms seemed to fit. "replace all" describes removing all the children, and replacing them with a single node, where the single node might be a fragment that then gets unpacked. But what we need is removing one particular child, and replacing it with its own content.

If we have, `<div>abc<b>def</b>ghi</div>`, and we want to block `b`, then we want `<div>abc def ghi</div>`. Replace all (with b's children) would deliver `<div>def</div>`.

PR #180, opened 2022-10-19
Add all known aria- attributes to known, baseline, and defaults (fix #177)

### LINKED ISSUE #177: Ensure that all existing `aria-` elements are known and allowed
I'll make a list of all existing `aria-` attributes. We do not want to remove accessibility features from HTML.

Sidenote: We do not have the same issue with `data-` attributes. Those attributes are removed by default and can be kept with a configuration as they will not be listed in `known` attributes. For `aria`, we want to keep them by default.

mozfreddyb:


otherdaniel (approved):
Thanks, looks good!

Issue #181, opened 2022-11-16
New WebIDL syntax with explicit name/namespace


evilpie:
We have been working on a new syntax that doesn't involve any micro syntax for namespaces, but instead either supports just a string (with an implicit default namespace) or an object/dictionary with an explicit name and namespace.

```idl
dictionary ElementWithNamespace {
  required DOMString name;
  required DOMString _namespace;
}
typedef (DOMString or ElementWithNamespace) SanitizerElement;

enum Star {
  "*"
};
typedef (Star or sequence<SanitizerElement>) ElementsOrStar;

dictionary SanitizerAttribute {
  required DOMString name; // We aren't sure how to specify |any| attribute yet.
// maybe: optional bool defaultAttribute;
  DOMString? _namespace = null; // specifying null/undefined means "use the null namespace" 
  required ElementsOrStar elements;
}

dictionary SanitizerConfig {
  sequence<SanitizerElement> allowElements;
  sequence<SanitizerElement> blockElements;
  sequence<SanitizerElement> dropElements;
  sequence<SanitizerAttribute> allowAttributes;
  sequence<SanitizerAttribute> dropAttributes;
  // ....
}
```

## 1. Use case: Elements with no namespace (HTML)
```js
{
  allowElements: [ "div", "em", "span", "i"],
}
```

## 2. Use case: Mixing elements from different namespaces
```js
{
  allowElements: [ "div", "span", { name: "text", namespace:  "http://www.w3.org/2000/svg" }],
}
```
## 3. Use case: Attribute + elements
```js
{
  allowAttributes: [{name: "title", elements: ["i", "b", "s", "u"]}]
}
```
## 4. Use case: Allow some attribute on any element (e.g. global attributes)
```js
{
  allowAttributes: [{name: "title", elements: "*"}]
}
```

## 5. Use case: Allowing all attributes on a specific element
For this we aren't sure about the syntax. 
```js
{
  allowAttributes: [{name: "*", elements: ["span"]}],
  // or
  allowAttributes: [{defaultAttributes: true, elements: ["blah"]}) 
}
```

## 6. Use case: Non-null attribute namespace
```js
{
  allowAttributes: [{name: "href", namespace: "http://www.w3.org/1999/xlink", 
     elements: [{ name: "text", namespace: "http://www.w3.org/2000/svg" }]
  }]
}
```

mozfreddyb:
CC @annevk

domenic:
[...]

PR #182, opened 2022-11-22
Remove current namespace wording from the current draft.

otherdaniel:
Remove the current, prefix:name-based treatment of namespaces from the spec. The goal is to revert the spec to be HTML-only, so that we can then build up a new API along the lines of #181.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -870,6 +870,7 @@ Issue(146): Whitespaces or colons?
 To determine whether an <dfn>|attribute| matches an [=attribute match
 list=]</dfn> |list|, run these steps:
 
+  1. If |attribute|'s [=Attr/namespace=] is not `null`: Return `false`.
   1. If |attribute|'s [=Attr/local name=] does not match the
      [=attribute match list=] |list|'s
      [key](https://webidl.spec.whatwg.org/#idl-record) and if the key is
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -989,24 +876,10 @@ list=]</dfn> |list|, run these steps:
      not `"*"`: Return `false`.
   1. Let |element| be the |attribute|'s {{Element}}.
   1. Let |element name| be |element|'s [=Element/local name=].
-  1. If |element| is a in either the [=SVG namespace|SVG=] or
```
evilpie, line 992: I am wondering if we should explicitly `return false` for any _attribute_ that is not in the null namespace for now. (Similar to how we check for the HTML namespace of elements)

evilpie: Oh, I guess HTML elements never have namespaced attributes anyway.

evilpie:
LGTM

mozfreddyb:
Thanks for the review, @evilpie. Looks good to me too.

mozfreddyb (approved):
Forgot to click the button? 🤷 🙁 🙂

index.bs
otherdaniel: Yeah, that'd make sense.

(OTOH, I don't expect this wording to remain in the spec for long. I just want to remove the old text first.)

otherdaniel: Added a line that checks for attribute namespaces.

Issue #185, opened 2023-01-11
allowing unsafe markup


mozfreddyb:
We've been told that people **do** want to deviate from our security guarantees with all terrible, known consequences.

We already allow people to opt out of the "known elements" by switching `allowUnknownMarkup`. However, this still disallows e.g., `script` and friends.

Apparently, people may appreciated the sanitizer for it's abilities to modify and walk a DOM tree regardless of the security guarantees. This may necessitate the existence of a pref that allows arbitrary and insecure allow lists and just skips all list checks (baseline, known).

I'm suggesting we introduce `boolean allowUnsafeMarkup` in the SanitizerConfig dictionary.

Up until now this seemed very simple and straightforward but until today we have always allowed our friends working on Trusted Types to enjoy the shiny assumption that the Sanitizer does not need to be subject to TT checking & enforcement because of its strong security guarantees. 

With that ship sailing, we may have to revisit: Trusted Types would have to monkey-patch/check the SanitizerConfig.

We can still punt this to a v1.1.

CC @otherdaniel @annevk @koto

koto:
Perhaps not surprisingly, I don't think it's a good idea to allow for such setting :) 

It would directly contradict the [API goals](https://github.com/WICG/sanitizer-api#goals-for-the-sanitizer-api) (sanitizer with such config would not be safe). The explainer mentions:

> The Sanitzer is safe by default, which means it has built-in rules about which markup to keep or to discard. Developers can customize the Sanitizer to suit the needs of their applications. But the sanitization rules cannot be relaxed below a built-in, safe baseline configuration.

The proposed change in my opinion falls into [non-goals](https://github.com/WICG/sanitizer-api#non-goals) of the API:

> Some applications will have sanitization requirements that are not easily met by a general purpose library. These should continue to be able to use whichever library or mechanism they prefer. However, the library should play well with other enforcement mechanisms.

There are already several options available in the platform to parse HTML into a document, traverse a document, and mutate the nodes with arbitrary rules. There are existing libraries that do all of these, with user-provided config. The value add of the Sanitizer API is the security. Paraphrasing, I think Sanitizer API should at least sanitize ;)

benbucksch:
> people may appreciated the sanitizer for it's abilities to modify and walk a DOM tree

That suggests that there should be another API that can do this. And the sanitizer could use that API.

mozfreddyb:
[...]

PR #189, opened 2023-02-23
Recurse into shadow roots.

otherdaniel:
This proposes to recurse into any shadow roots.
This follows the discussion at https://github.com/WICG/sanitizer-api/issues/185

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -797,6 +797,9 @@ algorithm collects these in one place.
 <div algorithm="handle funky elements">
 To <dfn>handle funky elements</dfn> on a given |element|, run these steps:
 
+  1. If |element| is a [=shadow host=]:
+    1. Run the steps of [=sanitize a document fragment=] algorithm on
+       |element|'s [=Element/shadow root=].
   1. If |element|'s [=Element/namespace=] [=is=] [=HTML namespace|HTML=] and
      the [=Element/local name=] [=is=] `"template"`:
     1. Run the steps of the [=sanitize a document fragment=] algorithm on


index.bs
```diff
@@ -797,6 +797,9 @@ algorithm collects these in one place.
 <div algorithm="handle funky elements">
 To <dfn>handle funky elements</dfn> on a given |element|, run these steps:
 
+  1. If |element| is a [=shadow host=]:
+    1. Run the steps of [=sanitize a document fragment=] algorithm on
+       |element|'s [=Element/shadow root=].
```
annevk, line 802: I suggest adding a note here that this is okay to do since we know our only entry point is the fragment parser.

If we ever have another entry point, such as an arbitrary node, we need a more complicated handshake to not violate encapsulation.

annevk: Also, should this be part of "handle funky elements"? It doesn't matter much but it seems this applies to all elements. But maybe "handle funky element" should be "sanitize an element" and all element-logic should move here?

mozfreddyb: If you call into `sanitizer a document fragment`, you need to pass a `Sanitizer` instance (I think `this`).
I too think it might make sense to throw stuff into a "handle element" function. Otherwise looks good.

mozfreddyb:
I assume the spec is diverging from this section with your ongoing rewrite. OK to close?

otherdaniel:
> I assume the spec is diverging from this section with your ongoing rewrite. OK to close?

Yes to both.

Issue #192, opened 2023-05-03
2023-05-03 meeting notes


annevk:
- [Mozilla’s Sanitizer API Position document](https://docs.google.com/document/d/1AJgtEGZwHR24S5lW6Cin-dr-g_AoeVTsxUrECj34Pd0/edit)
- [Google’s Sanitizer API Position document](https://docs.google.com/document/d/1RDI-IIuu1Of95KON8Ox3yf-eUWmfyiS6wxeQRy_SxDY/edit)

General agreement on:
- Some string -> Document with a filter API (e.g., Document.parseUnsafe())
- Some string -> Document with a must-be-safe filter API (e.g., Document.parse())
- Some string -> parse + insert with a filter API (e.g. element.setHTMLUnsafe())
- Some string -> parse + insert with a must-be-safe filter API (e.g. element.setHTML())

Details to discuss:
- How are filters delivered? { filter } or { sanitizer } or { filter } for one and { sanitizer } for the  other?
- Are filters dictionaries or Sanitizer objects or either allowed or ...?
    - If there were to be a Sanitizer object it should be useful, e.g. with properties reflecting the "resolved" allowlists you could inspect.
- Should there be positional methods?
   - Previous discussion at HTML triage meeting: https://github.com/whatwg/html/issues/8942#issuecomment-1462965265

shgysk8zer0:
I am struggling to understand what changed in the new API design.

- I noticed all mentions of `Sanitizer` have been removed
- The [WICG unofficial draft](https://wicg.github.io/sanitizer-api/) (2023-05-31) is still mostly about the construction and methods, etc. of `Sanitizer`
- The [current explainer](https://github.com/WICG/sanitizer-api/blob/main/explainer.md) has no examples
- #194 Adds examples using `element.setHTML` & `Document.parseHTML` that seem to imply passing the options/filter directly to these methods (`element.setHTML(html, { allowElements, allowAttributes })`) instead of `element.setHTML(html, { sanitizer: new Sanitizer({ allowElements, allowAttributes}) })`
- Those examples also differ greatly in structure from the [default configuration dictionary](https://wicg.github.io/sanitizer-api/#default-configuration-dictionary) (possibly to address SVG and MathML)

I get the impression that `Sanitizer` has been removed and that `setHTML` & `parseHTML` are the direction this is heading, possibly passing in what was formerly the the arguments to `new Sanitizer(config)`. Is this correct?

annevk:
[...]

Issue #188, opened 2023-02-08
Clarify threat model, "XSS first and foremost"


otherdaniel:
This is a follow-up to https://github.com/WICG/sanitizer-api/issues/183. As I understand that issue, we would like to tighten the Sanitizer threat model to be specifically about XSS / script injection, with anything else being optional. This would have several consequences for the API in general:

Baseline and default would be XSS-focused, and would not be different concepts anymore. The distinction would just drop out.  https://github.com/WICG/sanitizer-api/issues/183 covers this.

Additional cleanup items:

- There are several config options - allowComments, allowCustomElements, allowUnknownMarkup - which do not seem to defend against XSS-y threats. In this sharpened threat model, they'd default to true?
- https://github.com/WICG/sanitizer-api/issues/185 discusses `allowShadowRoots`. It seems that discussion gravitates towards default-true anyhow, but.. that'd also be default-true?
- The current spec unconditionally drops some node types, like CDATA or PI, which I think are still relevant to XML documents. In or out?

annevk:
The sanitizer should take care of shadow roots. I think it makes sense for it to recurse into them. As there's currently no way to remove a shadow root from a shadow host.

Not dropping PI might allow XSLT (although that wouldn't work in subtrees).

otherdaniel:
Agree on shadow roots.

Also agree that XSL transformations are XSS-equivalent. As I read it, XSLT wants a PI with a target `xml-stylesheet` that is a child of the root node before the first element. I'd assume `setHTML` defined on the Element - or any other entry point that doesn't take a Document - wouldn't be affected.

annevk:
Yeah. I was thinking later on that we should "Assert" https://infra.spec.whatwg.org/#assertions that the node is not CDATA or PI as the HTML parser cannot return those and for now we'd only run on HTML parser output. When there is an endpoint that can contain those we can carefully evaluate again as we'd also need to consider some other angles at that point anyway.

That leaves your first bullet point and I'm less clear on that.

allowUnknownMarkup seems like a potential forward compatibility problem and as such probably safer to remove by default.

allowCustomElements sits between safe and unsafe. We don't know what the JS implementation will do and how vulnerable it is. I'd lean towards removing by default still.

allowComments was also safer to remove based on discussion I thought. In particular because they have been a common mutated XSS vector in the past.

otherdaniel:
[...]

Issue #183, opened 2022-12-14
`baseline` should be `default` and `default` should be `baseline`


mozfreddyb:
Most other sanitizers have a default list that is according to some standard (an opinionated default, a default that cares about XSS and nothing else, ...). Most other sanitizers also allow configurations in which developers may re-allow dangerous and XSS-y markup.

We aim for the Sanitizer API to not allow configurations that might reintroduce XSS. For that reason we've had a couple of "list categories" namely. We have one of each category for attributes and elements. The categories are as follows:
- baseline: The baseline lists for elements/attributes contain every known element that can not introduce XSS. Even if the element is able to lead to other attacks (form hijacking, DOM Clobbering etc.). This is our safety net. If you want to allow an element through the configuration, it *needs* to be in the baseline.
- default: These lists are an opinionated subset of the baseline, they existed with the intent of preventing some further attacks. We never fully specified which use cases we want to include or exclude in our default :(
- known: This is a last line of defense for configurations with `allowUnknownElements: true`. The intent is to allow configurations that include a `foobar` element which is obviously non-existant but also harmless. We check every unknown element to be *actually* unknown, as `allowUnknownElements: true` should not be used to re-introduce XSSy elements like `script`.

Lately, we've been discussing what the `default` is actually defending against and we could not really come up with a hard line.

My suggestion is to therefore remove the distinction between *baseline* and *default*. 
We should, however introduce threat-model specific configurations that are less "opinionated" and orient along hard facts. E.g., a configuration that disallows form hijacking & DOM Clobbering on top of the baseline.

In a background conversation, @otherdaniel was in favor of this but wanted to think about it some more before moving on.
We're curious to hear other opinions, but I hope not to dwell on this for very long.

benbucksch:
When I created the original sanitizer in Mozilla, it was for the purpose of Thunderbird sanitizing HTML emails. The goals were:
1. Remove any possibility at all to ever have a security hole. This specifically also included security holes not already known. The sanitizer was presuming that known holes would be fixed in the engine, but that other holes appear regularly (monthly), and the goal was stop those future holes as well. For that goal, any complex browser feature that is not absolutely necessary in email would be disabled.
  * Any and all forms of JS or other scripts had to be blocked, and any vector how scripts might be sneaking in. That was one of the primary goals of the sanitizer. This was a double-protection in addition to disabling JS in that iframe.
[...]

PR #193, opened 2023-05-04
First draft: explainer-new.

otherdaniel:
Explainer for new Sanitizer API design, based on recent sync meeting (#192).

Notes:
- Not ready yet; would like to add examples.
- I added several open questions not discussed in the meeting.


otherdaniel:
This tries to summarize the result from the recent sync meeting. I hope it's a starting point to replace the current explainer for this project, and to then base future work on it. Please let me know if this doesn't reflect the meeting results accurately.

I've added several open questions (i.e., what about defaults?) that were not explicitly discussed at the meeting.

I'd like to add examples, too, to make this more useful for people not involved in the meetings.

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+On one axis they differ in the parsing context and are analogous to `innerHTML`
+and `DOMParser.parseFromString`, respectively; on the other, they differ in
+whether they enforce an XSS-focused security baseline or not. These two aspects
+pair well and yield:
+
+- `Element.setHTML(string, {options})` - Parses `string` using `this` as
+  context element, like assigning to `innerHTML` would ; applies a filter,
+  while enforcing an XSS-focused baseline; and finally replaces the children
+  of `this` with the results.
+- `Element.setHTMLUnsafe(string, {options})` - Like above, but it does not
```
annevk, line 20: I didn't like "unsafe" at the end initially, but for sorting purposes it is kinda nice I suppose.

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+  enforce a baseline. E.g. if the filter is configured to allow a `<script>`
+  element, then a `<script>` would be inserted.
+- `Document.parseDocument(string, {options})` (static method) - Creates a new
```
annevk, line 23:
```suggestion
- `Document.parseHTML(string, {options})` (static method) - Creates a new
```
`Document` is redundant and as we'll only define HTML code paths for now the name should reflect that. Same below.

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+- Enforcement of a security baseline moves to the method. The filter config -
+  formerly the sanitizer config - can now be used in differently,
+  either in a guaranteed-secure way or in use-config-as-written way.
+- Since the 'sanitizer' config can now be used in both safe and unsafe ways,
+  it's arguably no longer a sanitizer config but a filter config.
+
+## Open questions:
+
+- Defaults: If no filter is supplied, do the safe methods have any filtering
+  other than the baseline? Do the unsafe methods have default filtering?
```
annevk, line 46: 1. It seems good to conclude #188 so "baseline" is clear.
2. I think that by default "unsafe" shouldn't do any filtering. And if you want you get to essentially supply your own block and safelists and the browser won't perform any normalization on them.

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+  (I.e., would they allow `<script>` by default, or would they drop `<script>`
+  by default but allow it if specifically allow-listed.)
+- Defaults: All of these are new methods, without legacy usage. Would DSD
+  parsing default to `true`? Do we even decide that, or would we instead ask
+  the HTML WG and adopt whatever their choice?
```
annevk, line 51: I think ultimately that decision is largely up to the WHATWG. And we decided there previously that new parsing APIs won't need any kind of opt-in for declarative shadow roots. So we shouldn't add any here. I don't think we need opt-out either (that's handled through custom elements not being safelisted).

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+- Should the filter config be a separate object, or should it be a plain dictionary?
+  - Reasons pro dictionary:
+     - Simpler.
+  - Reasons pro object:
+     - Allows to pre-process of the config and to amortize the cost over many calls.
+     - Allows adding other useful config operations, like introspection.
```
annevk, line 57: If actual normalization is performed on the input and that is exposed (rather than the input being reflected back) I could see folks supporting a class. The current `Sanitizer` class is not it though.

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+- Exact filter options syntax. I'm assuming this will follow the discussion in
+  #181.
+- Naming is TBD. I'm trying to follow the preferences from the recent 'sync' meeting.
+- Baseline for safe methods: There are some "special" behaviours (currently in
+  "handle funky elements"), mainly around dropping javascript:-URLs in contexts
+  that navigate. Should this be available as an option (which would then be
+  force-true for safe usage, and default-false but available for non-safe
+  usage), or would that be custom behaviour only for the safe methods?
```
annevk, line 65: I'd be okay leaving it out until we see demand.

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+- Should the `DOMParser`-like methods offer a mimetype options, like DOMParser
+  does?
```
annevk, line 67: No, that option boils down to a boolean and I think we settled on incorporating that boolean into the method name (at least we did for `setHTML()` so it makes sense to follow that).

annevk (commented):
Thanks for writing this up @otherdaniel! Left my thoughts and nits inline.

explainer-new.md
mozfreddyb: As long as a declarative shadow root always requires a custom element as a parent and as long as we require custom elements to be named in the `allowElements` list, I am very OK with this not being a toggle at all.

explainer-new.md
mozfreddyb: Yes, I'm +1 for doing this HTML-only and name the methods accordingly.

explainer-new.md
mozfreddyb: 1. This is more than "what the baseline is", but also whether the default should be something "more usable" than the baseline (e.g., also issue #183). I think it should suffice to enforce only the baseline. (Unless we actually do the full research on the defaults of other sanitizers and e.g., popular XSS gadget attributes to also strip from the "default".)

explainer-new.md
mozfreddyb: My main hope with this was that a framework would supply a Sanitizer instance for the web developer to use in framework-specific code (e.g., to prevent script gadgets attacks). That being said, a dictionary is likely as good.

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+  the HTML WG and adopt whatever their choice?
+- Should the filter config be a separate object, or should it be a plain dictionary?
+  - Reasons pro dictionary:
+     - Simpler.
+  - Reasons pro object:
+     - Allows to pre-process of the config and to amortize the cost over many calls.
+     - Allows adding other useful config operations, like introspection.
+- Exact filter options syntax. I'm assuming this will follow the discussion in
+  #181.
+- Naming is TBD. I'm trying to follow the preferences from the recent 'sync' meeting.
```
mozfreddyb, line 60: Sorry, bike shed. Not super happy with filter, reminds me of `NodeFilter` and that's some hairy stuff. Let's keep this in mind but postpone the discussion as to not block updating the explainer any further.

explainer-new.md
```diff
@@ -0,0 +1,72 @@
+# Sanitizer API: New API Explainer
```
mozfreddyb, line 1: Nit: Can/Should we replace `explainer.md` instead of setting a new one? It will be available through git history and I don't think we should need it.

mozfreddyb (commented):
This is great. Thanks so much, Daniel!
(And my deepest apology for the the review delay here!)

I think this is almost ready to merge once we have resolved the minor sidebars

explainer-new.md
```diff
@@ -0,0 +1,72 @@
...
+Context: Various disagreements over API details lead to a re-design of the
+Sanitizer API. The current (April '23) spec draft no longer represents a
+consensus within the Sanitizer working group (or other interested groups, like
+the HTML WG). This explainer sketches out a new API design:
+
+## A new proposal, based on #192:
+
+There is a 2x2 set of methods that parse and filter the resulting node tree:
+On one axis they differ in the parsing context and are analogous to `innerHTML`
+and `DOMParser.parseFromString`, respectively; on the other, they differ in
```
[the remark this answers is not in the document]
otherdaniel: Done.

explainer-new.md
otherdaniel: Done.

explainer-new.md
otherdaniel: Done. (Removed this from open questions, and added a sentence after the API proposal.)

explainer-new.md
otherdaniel: Reworded this. Agree on WHATWG getting the last word.

I'm not aware of any requirements that shadow roots - declarative or not - must be hosted by custom elements. I thought a boring old `<div>` could also host a shadow DOM. Am I overlooking something?

annevk (commented):
Thanks, this looks great!

explainer-new.md
mozfreddyb: Well, glad we got some clarity here. I was totally not aware that this may live outside of custom elements. That being said, I think this is all fine as long as the sanitizer traverses the shadow root.

explainer.md
```diff
@@ -54,119 +54,72 @@ a general purpose library. These should continue to be able to use whichever
...
+There is a 2x2 set of methods that parse and filter the resulting node tree:
+On one axis they differ in the parsing context and are analogous to `innerHTML`
+and `DOMParser`'s `parseFromString()`; on the other axis, they differ
+in whether they enforce an XSS-focused security baseline or not. These two
+aspects pair well and yield:
+
+- `Element.setHTML(string, {options})` - Parses `string` using `this` as
+  context element, like assigning to `innerHTML` would; applies a filter,
+  while enforcing an XSS-focused baseline; and finally replaces the children
+  of `this` with the results.
```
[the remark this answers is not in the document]
mozfreddyb: While technically correct, I think it helps to draw a comparison. Even if it's not exactly 1:1, I am expecting many developers to switch from `innerHTML=` to `setHTML()`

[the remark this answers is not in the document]
annevk: This is summarizing the design for implementers, no? I don't mind it saying `innerHTML`, but it should be more accurate.

explainer-new.md
mozfreddyb: We discussed this on the call and agree that this is fine as long as the sanitizer is traversing

README.md
```diff
@@ -7,135 +7,13 @@ Status:
...
-* The [draft specification](https://wicg.github.io/sanitizer-api/).
-* A list of [questions & answers](faq.md).
-* [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API).
-* Implementation Status:
-  * [Mozilla position](https://github.com/mozilla/standards-positions/issues/106),
-    [Chrome Status](https://www.chromestatus.com/feature/5786893650231296),
-    [WebKit position](https://lists.webkit.org/pipermail/webkit-dev/2021-March/031738.html).
-  * [Can I use 'Sanitizer API'](https://caniuse.com/mdn-api_sanitizer)?
-  * [Web Platform Tests](https://wpt.fyi/results/sanitizer-api?label=experimental&label=master&aligned)
-    ([test source](https://github.com/web-platform-tests/wpt/tree/master/sanitizer-api)).
```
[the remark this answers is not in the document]
otherdaniel: Done.

explainer.md
```diff
@@ -54,119 +54,72 @@ a general purpose library. These should continue to be able to use whichever
...
+- Enforcement of a security baseline depends on the method. The filter/sanitizer
+  config can now be used differently, either in a guaranteed-secure way or in
+  use-config-as-written way.
+
+## Open questions:
+
+- Defaults: If no filter is supplied, do the safe methods have any filtering
+  other than the baseline?
+- Defaults: All of these are new methods without legacy usage. Would DSD
+  parsing default to `true`? (Probably. Decision lies with WHATWG.)
```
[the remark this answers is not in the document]
otherdaniel: Removed. (I added this as one example to the note about differences between the new APIs and their existing counterparts.)

explainer.md
```diff
@@ -54,119 +54,72 @@ a general purpose library. These should continue to be able to use whichever
...
+- Should the filter config be a separate object, or should it be a plain dictionary?
+  - Reasons pro dictionary:
+     - Simpler.
+  - Reasons pro object:
+     - Allows to pre-process the config and to amortize the cost over many calls.
+     - Allows adding other useful config operations, like introspection.
```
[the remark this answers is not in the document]
otherdaniel: Done. Adopted this wording.

explainer.md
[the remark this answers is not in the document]
annevk: I meant it for both.

mozfreddyb (approved):

annevk (approved):
\o/

Thanks!

Issue #101, opened 2021-06-18
Error handling for string functions when parsing context element would be blocked by config.


otherdaniel:
I'm presently mucking around with an implementation of the API along the lines of https://github.com/WICG/sanitizer-api/issues/42#issuecomment-857757109 , to see whether it "feels" right. It mostly does, except for error handling: Since we now have an explicit context element... should the config apply to the context or not?

The issue comes up since several fairly simple test cases now break: There's a new test cases along the lines of:
```js
  new Sanitizer({allowElements: ["p"]}).sanitize("<p>bla<em>bla</p>")  // "<p>blabla</p>
```

For re-writing the tests, I now need to pick a context (regardless of whether directly in the API, or indirectly by parsing separately). Most test don't use anything fancy, so I thought I'd just pick a `<div>` or `<template>`. In my current implementation, I'm also checking the context against the configuration (built-in & supplied), which means that neither context would work, since I'm only allowing `<p>`. That semms logical, but... fairly annoying. In particular, it means that you e.g. cannot sanitize into a `<template>` element without also allowing `<template>` in the tree.

On the other hand, if we would *not* check the context element against the configuration, then you could do something like `.sanitizeFor("script", "alert(1)")`, which seems like even worse.

```js
  // Should this be: <template><p>blabla</p></template>, or null?
  new Sanitizer({allowElements: ["p"]}).sanitizeFor("template", "<p>bla<em>bla</p>")

  // Is this different from:
  ... .sanitizeFor("script", ...)
  some_script_element.SetInnerHTML("alert(1)") 
```

What I have presently implemented:
- The context element gets checked against the configuration. (And thus also against the baseline & defaults).
- `.sanitizeFor` returns null if the context would have been removed.
- `.SetSanitizedHTML` / `SetHTML` throws if the context element would have been removed (since it can't fullfil the contract.)

Solutions I'm seeing:
1. Check the context, as described above.
2. Never check the context.
3. Check the context, but with a more restricted check. (E.g. only against the baseline.)
4. ???

mozfreddyb:
We've discussed this in the meeting. We're leaning towards ignoring the context element as part of the sanitizer pass. Essentially, we'd parse the input into a document fragment and only perform the sanitizer pass only on that resulting fragment.

Also, this would be inconsistent if we'd pay attention to a context element but not it's attributes.

@otherdaniel  is going to think this through and incorporate this into e.g. #196.

otherdaniel:
[...]

PR #196, opened 2023-07-14
Add more examples.

otherdaniel:
This adds additional examples to the explainer and largely reaches "feature parity" with the existing spec, but based on our current thinking.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -96,6 +96,8 @@ that navigate.
 The 'unsafe' methods will not apply any filtering if no explicit config is
 supplied.
 
+> [!Note] The 'unsafe' methods are being worked on here: https://github.com/whatwg/html/pull/9538
+
 ## Major differences to previously proposed APIs:
 
 The currently proposed API differs in a number of aspects:
@@ -106,6 +108,7 @@ The currently proposed API differs in a number of aspects:
 - Enforcement of a security baseline depends on the method. The filter/sanitizer
   config can now be used differently, either in a guaranteed-secure way or in
   use-config-as-written way.
+- The configuration dictionary differs substantially in syntax.
 
 ## Open questions:
 
@@ -115,10 +118,6 @@ The currently proposed API differs in a number of aspects:
   dictionary? (As-is, it should probably be a dictionary. An object would
   require either compelling performance numbers, or a compelling operation that
   would only work with a pre-processed dictionary.)
-- Exact filter options syntax. I'm assuming this will follow the discussion in
-  #181.
-- Naming is TBD. Here I'm trying to follow the preferences expressed in the
-  recent 'sync' meeting.
 
 ## Examples
 
@@ -148,6 +147,8 @@ Document.parseHTML(example_tr);  // <html><head></head><body>A table row.</body>
 All of these would have had identical results if the "unsafe" variants had
 been used.
 
+### Parsing XML
+
 Parsing follows HTML parsing rules, unlike `innerHTML`, where it depends on the
 document type:
 ```js
@@ -161,6 +162,8 @@ element_xml.setHTML(example_not_xml);  // <div xmlns="http://www.w3.org/1999/xht
 element.setHTML(example_not_xml);  // Same as above.
 ```
 
+### Safe vs Unsafe methods
+
 The "safe" methods remove all script-y content defined by the platform and
 let the rest pass:
 ```js
@@ -168,32 +171,33 @@ element.setHTML(`<a href=about:blank onclick=alert(1) onload=alert(2) id=myid cl
 // <div><a href="about:blank" id="myid" class="something"></a></div>
 ```
 
+### Configuration Options: Basic use and namespaces
+
 The operation of the built-in sanitizer can be configured to suit your
 applications' needs. Both "safe" and "unsafe" versions can take a configuration.
-(Please note that naming and structure here is rather preliminary,
-but we expect these capabilities to be in the final standard.)
 
 The "safe" version will ignore configuration items that break its security
 guarantees:
 ```js
-const an_unsafe_config = { 'allowElements': [ { name: 'script' } ] };
+const an_unsafe_config = { 'elements': [ { name: 'script' } ] };
 element.setHTML("<script>", { sanitizer: an_unsafe_config });  // <div></div>
 element.setHTMLUnsafe("<script>", { sanitizer: an_unsafe_config });  // You now have a script. Congrats.
 ```
 
-For elements, the HTML namespace is default. For attributes, the null namespace.
+For elements the HTML namespace is default. For attributes, the null namespace.
 Other namespaces can be supported. A string entry stands for a dictionary with
 only the name, in the HTML/null namespace (for elements/attributes,
 respectively).
+
 ``` js
 const config_with_namespaces = {
-  allowElements: [
+  elements: [
     'a',  // The HTML anchor element.
     { name: 'a' },  // Also the HTML anchor element.
     { name: 'a', namespace: 'http://www.w3.org/1999/xhtml' },  // Another one.
     { name: 'a', namespace: 'http://www.w3.org/2000/svg' }  // SVG's anchor element
   ],
-  allowAttributes: [
+  attributes: [
     'href',  // An href attribute. The one you'd expect on an HTML anchor.
     { name: 'href' },  // The very same.
     { name: 'href', namespace: '' },  // There it is again.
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -215,11 +224,189 @@ defaults.
...
+be rewritten as an allow-lists of equal or shorter length by removing any
+block-list element from the allow-list, or dropping it altogether if there
+was no matching allow-list entry in the first place.
+
+```js
+// equivalent to:  { allowElements: ["i"] }
+const config_that_mixes_allow_and_block_lists = {
+    allowElements: ["i", "u"],
+    blockElements: ["u", "s"],
+};
```
mozfreddyb, line 294: That's all correct and matches my understanding. I think it would make more sense for a reader if we defined and then pointed to the processing model (Roughly "check first drop / block first, then allow." like we already have in the "determine the sanitize action for an element" algorithm)

explainer.md
```diff
@@ -215,11 +224,189 @@ defaults.
...
+const drop_src_attribute_from_input = {
+  dropElements: [{ name: "input", attributes: ["src"]}],
+}
+```
+
+TODO: What does `{ name: "bla", attributes: [] }` mean? My gut feeling would
+  be that `attributes: []` is a no-op and it allows `<bla>` plus whatever
+  attributes are allowed globally. But if so, the only way to describe an
+  element without any attributes in presence of a global `allowAttributes`
+  would be to add a dropElements for it.
```
mozfreddyb, line 349: To be clear, this question is for `allowAttributes: { name: "bla", attributes: [] }` right?
Hm, what if empty actually meant "no attributes for this element"? That would make the list look-up "just work" and allow special-casing elements where the global `allowAttributes` shouldn't apply (I am assuming both rules must allow the attribute).

explainer.md
```diff
@@ -215,11 +224,189 @@ defaults.
...
+
+TODO: I find the syntax weird. In the allowElements case, the `img` entry
+  matches (and hence allows) both the element and the attribute. In the drop
+  case it matches only the attribute (but not the element). If we were to
+  define this identically in both cases - always match element and attributes -
+  then it simply wouldn't work for drop, because a drop-rule with attributes
+  would automatically drop the element (and hence all attributes with it).
```
mozfreddyb, line 356: Can you expand? I'm not sure I fully understand the example again. 
My understanding:
- We go through all elements and if it is allowed, the list of attributes
- We go through all attributes and keep them if allowed

These "keep if allowed" steps would go through drop/block/allow for elements and block/allow for attributes in that order. If all of those 3 (for elements) or 2 (for attributes) lists allow the node, then it should be kept.

If the attribute-list is a list of attributes, we would match as "the attribute is in list" just according to the attribute. If the attribute-list is a list of attributes+elements, then it would only match as "is in list" if both element & attribute fit.

Does that make sense? Did I misunderstand the example?

explainer.md
```diff
@@ -215,11 +224,189 @@ defaults.
...
+
+How do multiple attribute rules interact? An attribute is allowed, if at
+least one allow-rule (global or per element) allows it, and no drop
+rule drops it.
```
mozfreddyb, line 360: I think we would go through the lists in order. If it is dropped, then we don't continue looking into the allow list. Similar to existing spec text. No?

mozfreddyb (approved):
Looks good! Let's make sure we have a shared understanding for the TODOs before we merge.

explainer.md
otherdaniel: I meant `allowElements`, not `allowAttributes`.

Given the input `""`, what do these do:

* `allowElements: [{ name: "img", attributes: [] }]`
* `allowElements: ["img"]`
* `allowElements: `[{name: "img", attributes: undefined }]`

My gut feeling is that these three should mean the same thing. (I could certainly be convinced otherwise.)

On: _"allow special-casing elements where the global `allowAttributes` shouldn't apply (I am assuming both rules must allow the attribute)."_

* Would `allowElements: [{ name: "img", attributes: [] }], allowAttributes: ["src", "srcset"]` turn the input example above into "", because the attributes are not allowed by both rules?

I'm worried that we're making the "grab bag" of elements and attributes very laborious, which was the primary use case we found when looking at existing sanitizer usage.

explainer.md
otherdaniel: What does `blockElements: [{ name: "img", attributes: ["srcset"] }]` mean? Does it mean - given the sample input from the earlier comment - an empty tree, or ``? If the former, why have the `attributes:` key at all? If the latter, then what does `blockElements: [{ name: "img", attributes: [] }]` mean?


The currently proposed syntax allows multiple clauses for one element. What do these mean:

* `allowElements: [{ name: "img", attributes: ["src"]}, { name: "img", attributes: ["srcset"]}]`
* `blockElements: [{ name: "img", attributes: ["src"]}, { name: "img", attributes: ["srcset"]}]`

(Btw, I'm not complaining; I'm just trying to sound out the edge cases.)

explainer.md
mozfreddyb: 1. Yes, I think all of those three examples should resulti n `` with no attributes.
2. Great point. We shouldn't make this too cumbersome, I agree. I guess this could work if our step for "finding the allowed attributes for this current element" were to build the union of what `allowElement` and `allowAttributes` are specifying. But then, how would you build an instruction that is "_allow an element `img` with no attributes at all, not even the ones specified elsewhere_". Can't think of something but happy to go with a suggestion from you.

explainer.md
mozfreddyb: From the top of my head (still not fully back at my desk mentally), I'm not sure what the attribute-pieces even means for the `blockElements` case. Block the element if it has those attributes? Block the attributes but only on these elements? I guess the latter..

explainer.md
mozfreddyb: Discussion from today's meeting to explain the issue at hand a bit better.

Example: allowing `{name: "div", attribute: "foo"}, {name: "div", attribute: "bar"}`and input `<div foo="" bar="">`

If the algorithm would go through each entry individually, leading to `<div>`.
If the algorithm would collect all entries and combine it would return the input.

explainer.md
```diff
@@ -215,11 +224,206 @@ defaults.
 ```js
 const config_allow = {
   allowElements: [ "div", "p", "em", "b" ]  // Allows only those four elements.
-      // Output with "safe" and "unsafe" methods should be the same.
+  allowAttributes: [ "class" ]  // Allows only class attributes.
+      // Output with "safe" and "unsafe" methods are the same for this config.
 };
 const config_block = {
-  blockElements: [ "style" ]  // Allows a lot of things. But not <style>.
+  blockElements: [ "style" ],  // Allows a lot of things. But not <style>.
```
mozfreddyb, line 231: s/Allows a lot of things/Allows the default, except <style>/, maybe?

explainer.md
```diff
@@ -215,11 +224,206 @@ defaults.
...
+### Configuration List-Types: Allow-lists, Blocks-lists, and Drop-lists
+
+For elements there are three types of lists: Allow-lists, Block-lists
+and Drop-lists. These correspond to three related but somehow different use
+cases:
+
+If you have a specific set of functionality in mind that you wish to
+retain and would like to drop everything else as a matter of caution, this
+can be accomplished with allow lists. This is probably the most common case.
+For example, you might want allow user comments with simple styling, but
```
mozfreddyb, line 247: Nit: they are called "formatting elements"

explainer.md
```diff
@@ -215,11 +224,206 @@ defaults.
...
+```
+
+If you want to drop `src` attributes from `<input>` elements but retain them
+elsewhere, you can use:
+
+```js
+const drop_src_attribute_from_input = {
+  dropElements: [{ name: "input", attributes: ["src"]}],
+}
+```
```
mozfreddyb, line 308: Should this drop the element or the attribute?

explainer.md
```diff
@@ -215,11 +224,206 @@ defaults.
...
+
+An element with no allowed attributes follows from these definitions. For an
+element that allows any of the default-allowed attributes, you can use the
+string `"*"`.
+
+```js
+const config_div_without_anything = {
+  allow_elements: [ { name: "div", attributes: [] } ]
```
mozfreddyb, line 316: nit: s/allow_elements/allowElements/

explainer.md
```diff
@@ -215,11 +224,206 @@ defaults.
...
+};
+const config_div_with_everything = {
+  allow_elements: [ { name: "div", attributes: "*" } ]
+};
+```
+
+### Matching any attribute on a given element
+
+For some use cases, one may wish to allow (or drop) any attribute:
```
mozfreddyb, line 325: same question as above

explainer.md
```diff
@@ -215,11 +224,206 @@ defaults.
...
+    in the drop-list preserve their child contents, while the allow-list does
+    not.
+* Any allow-, drop-, or block-list should contain each name at most once.
+  * This would apply to short forms as well.
+    E.g., `["div", { name: "div", namespace: "http://www.w3.org/1999/xhtml" }]`
+    would throw.
+  * While lists with duplicate element or attribute names could be coalesced,
+    it is ambiguous what the meaning of duplicate elements with different
+    element-dependent attribute lists would be.
+* The name must be set. If an `attributes` key is present, it must be non-empty.
```
mozfreddyb, line 393: non-empty? We have "empty array" examples above. Maybe disambiguate

mozfreddyb (commented):
Thanks a lot for writing these cases down! Some questions above.

explainer.md
mozfreddyb: Looking back at https://github.com/WICG/sanitizer-api/issues/181, we had `dropElements` just accepting elements and `dropAttributes` to accept `{name: 'src', elements: ['input'] }`, which is a bit inverse of what the allow-properties do

explainer.md
```diff
@@ -215,11 +224,206 @@ defaults.
...
+In the example `config_text_with_only_simple_styling` in the previous chapter
+we have allowed the `src` attribute on any of the eight allowed elements.
+If one wanted to allow `class` everywhere, but `src` only on `<img>`, the
+following would do:
+
+```js
+const config_with_element_specific_attributes = {
+  allowElements: [
+    "p", "b", "i",
+    { name: "img", attributes: [ "src" ] }
```
mozfreddyb, line 295: `allowElements` is just elements.

explainer.md
```diff
@@ -205,21 +209,213 @@ const config_with_namespaces = {
...
+const config_allow_some_formatting = {
+  elements: [ "div", "p", "em", "b", "img" ],  // Allows only 5 elements.
+  attributes: [ "class" ]  // Allows only class attributes.
+      // Output with "safe" and "unsafe" methods are the same for this config.
 };
-const config_block = {
-  blockElements: [ "style" ]  // Allows a lot of things. But not <style>.
+const config_anything_but_style_definitions = {
+  removeElements: [ "style" ],  // Allows a lot of things. But not <style>.
+  removeAttributes: [ "class", "style" ]  // No style or class attribute either.
```
mozfreddyb, line 234: Nit: maybe instead of "anything" or "a lot of things", let's say "the defaults except ..."

explainer.md
```diff
@@ -205,21 +209,213 @@ const config_with_namespaces = {
...
+    in the flatten list preserve their child contents, while the allow-list does
+    not.
+* Any allow-, remove-, or flatten-list should contain each name at most once.
+  * This would apply to short forms as well.
+    E.g., `["div", { name: "div", namespace: "http://www.w3.org/1999/xhtml" }]`
+    contains the same name twice and would thus throw.
+  * While lists with duplicate element or attribute names could be coalesced,
+    it is ambiguous what the meaning of duplicate elements with different
+    element-dependent attribute lists would be.
+* The name must be set. If an `attributes` or `removeAttributes` key is present,
```
mozfreddyb, line 370: Nit: s/name/element name/

explainer.md
```diff
@@ -205,21 +209,213 @@ const config_with_namespaces = {
...
+element.setHTML("bla", {sanitizer: config_with_dupes});  // throws.
+
+const config_with_dupes2 = {
+  elements: [
+    { name: "div", attributes: ["class"] },
+    { name: "div", attributes: ["style"] }
+  ] };
+element.setHTML("bla", config_with_dupes2);  // throws.
+
+// Undefined attributes. What does it mean?
```
mozfreddyb, line 404: Is that a WebIDL questions?
I don't even know where to look that up... @annevk can you help?

mozfreddyb (approved):

explainer.md
otherdaniel: This was meant as applying to both element and attribute lists, so that a dictionary without a name-key set would be invalid.

explainer.md
otherdaniel: Done.

explainer.md
otherdaniel: It was meant to highlight that `{ name: "div", attributes: undefined }` could mean the same as either `{ name: "div" }` (allows the default attrs for `div`) or `{ name: "div", attributes: [] }` (allows no attrs for `div`). At least to me it's not obvious what interpretation to pick.

But.. you're right. Maybe WebIDL solves that problem for us and interprets in a way that does make it clear. I didn't think of that.

explainer.md
mozfreddyb: Fair enough. Keep it :)

annevk (approved):
Thanks so much for the quick turnaround @otherdaniel!

This looks good and I'm happy to see this merged with these caveats:

* We'll need to continue to discuss the "default" configuration and how it "combines" with a userland configuration.
* I'm not sure an option around shadow roots is workable per se. It might have to be a shadowHost option that you can set to "keep" (default), "remove", or "flatten", but even that could be problematic if we ever offer this outside of these parser contexts as we don't want you to be able to remove shadow roots of arbitrary elements.

explainer.md
```diff
@@ -205,21 +209,213 @@ const config_with_namespaces = {
...
+chiefly useful to remove unwanted formatting from user input, while
+preserving its textual content.
+
+```js
+const config_that_removes_elements_but_preserves_their_text_content = {
+ flattenElements: ["span", "em", "u", "s", "i", "b"]
+};
+
+element.setHTML(
+  "Fancy <b>text</b> with <span style='color:blue'>pizzazz</span>.",
```
neuracr, line 250: Would adding an example that flattens only one level help here?
e.g. 
```
const config_flatten_spans = {flattenElements: ["span"]};
element.setHTML("Fancy <span style='color:blue'> text with <b>pizzazz</b></span>.", { sanitizer: config_flatten_spans});
  // <div>Fancy text with <b>pizzazz</b>.</div>
```

explainer.md
```diff
@@ -205,21 +209,213 @@ const config_with_namespaces = {
...
+throw a `TypeError` instead of making a best effort attempt at interpreting
+the configuration. A well-formed configuration has the following properties:
+
+
+* It contains either an allow-list or a remove-list, but not both.
+  * This applies to both element and attribute lists, seperately.
+  * Note that any config with both, an allow-list and remove-list, can be
+    rewritten by removing the remove-list items from the allow-list and then
+    droping the remove-list entirely.
+  * A config with an allow-list and a flatten-list makes sense since the items
```
neuracr, line 360: What about flatten-list + block-list? I think it also makes sense.

1. When passing an allow-list, the implicit default is to block, so passing a block-list doesn't make sense.
2. When passing a block-list, the implicit default is to allow.
3. When passing both, the implicit default is to block, but then it doesn't make sense to define the block-list (back to point 1).

In all cases, passing a flatten-list makes sense because it's changing the implicit default.

explainer.md
```diff
@@ -205,21 +209,213 @@ const config_with_namespaces = {
...
+    in the flatten list preserve their child contents, while the allow-list does
+    not.
+* Any allow-, remove-, or flatten-list should contain each name at most once.
```
neuracr, line 363: Each name must not appear in a given list (allow-, flatten-, remove-) several times, but the same name shouldn't appear in several list either. Is this missing?

otherdaniel: True. I adapted the description.

explainer.md
```diff
@@ -205,21 +209,213 @@ const config_with_namespaces = {
...
+  * This would apply to short forms as well.
+    E.g., `["div", { name: "div", namespace: "http://www.w3.org/1999/xhtml" }]`
+    contains the same name twice and would thus throw.
+  * While lists with duplicate element or attribute names could be coalesced,
+    it is ambiguous what the meaning of duplicate elements with different
+    element-dependent attribute lists would be.
+* The name must be set. If an `attributes` or `removeAttributes` key is present,
+  it must be non-empty. It may contain an empty list, though.
```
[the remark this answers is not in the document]
otherdaniel: Yes. Removed it.

explainer.md
otherdaniel: Thanks. Removed the example.

explainer.md
```diff
@@ -205,21 +209,182 @@ const config_with_namespaces = {
...
+
+There is no `flattenAttribute` because attribute nodes do not have children.
+
+Flatten applies to its immediate children, i.e. to one level. Combining
+`allowElements` with `flattenElements` lets you keep some formatting, but all
+the text content:
+
+```js
+const config_flatten_spans = {
+  allowElements: ["b", "i"],
```
annevk, line 263:
```suggestion
`elements` with `flattenElements` lets you keep some formatting, but all
the text content:

```js
const config_flatten_spans = {
  elements: ["b", "i"],
```

explainer.md
```diff
@@ -205,21 +209,213 @@ const config_with_namespaces = {
...
+
+
+* It contains either an allow-list or a remove-list, but not both.
+  * This applies to both element and attribute lists, seperately.
+  * Note that any config with both, an allow-list and remove-list, can be
+    rewritten by removing the remove-list items from the allow-list and then
+    droping the remove-list entirely.
+  * A config with an allow-list and a flatten-list makes sense since the items
+    in the flatten list preserve their child contents, while the allow-list does
+    not.
```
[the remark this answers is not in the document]
otherdaniel: :-)  I can no longer figure out what I was actually trying to say there... I removed that sentence in favour of "Both allow-lists and remove-lists can be combined with flatten-lists."

explainer.md
```diff
@@ -148,6 +147,8 @@ Document.parseHTML(example_tr);  // <html><head></head><body>A table row.</body>
 All of these would have had identical results if the "unsafe" variants had
 been used.
 
+### Parsing XML
```
zcorpan, line 150: Should this say "XML documents" instead, since `setHTML` parses as HTML?

annevk:
I don't see #101 addressed here. @otherdaniel is that intentional?

PR #201, opened 2023-11-03
Fix the link markup

stof:


stof:
Note that some other links are broken a bit below: 
image

Those use a link syntax that is valid in markdown but may be unsupported in your bs markup language (I suspect that it might not like the inline markup inside the link text)

mozfreddyb:
(waiving the missing IPR check, due to this being a purely syntactical issue)

PR #211, opened 2024-03-22
Introduce a query-able Sanitizer (or config) object.

otherdaniel:
This makes a configuration query-able, and should make it easy (or easier) to derive a new configuration from an existing one. In terms of spec & implementation, this leans heavily on the existing canonicalization.

----

This needs further discussion. We had discussed re-introducing a configuration object in order to expose the canonical configuration(s), but not how exactly that should look like. This proposes specific spec & explainer wording to focus the discussion.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -221,6 +221,7 @@ page to query and predict what effect a given configuration will have, or
 to build a new configuration based on an existing one.
 
 <pre class=idl>
+[Exposed=Window,Worker]
 interface Sanitizer {
   constructor(optional SanitizerConfig config = {});
   SanitizerConfig get();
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -193,13 +194,69 @@ The <dfn for="DOM/Document">parseHTML</dfn>(|html|, |options|) method steps are:
...
 </div>
 
+## SetHTML options and the configuration object. ## {#configobject}
+
+The family of {{Element/setHTML()}}-like methods always take an options
+dictionary. Right now, only one member of this dictionary is defined:
+
+<pre class=idl>
+dictionary SetHTMLOptions {
+  Sanitizer? sanitizer;
```
annevk, line 211: This should not be nullable. And it should accept a dictionary directly as well. (Prolly needs to default to the empty dictionary therefore.)

annevk (commented):
Overall this shape looks good to me. Thanks for working it out. I kinda like not pre-committing to safe/unsafe, although it will be a challenge to optimize that.

mozfreddyb (approved):
LGTM, with a +1 to @annevk's comment above.

index.bs
otherdaniel: Done. (Also added a comment to the explainer.)

PR #212, opened 2024-03-28
Update "javascript:" handling; esp. SVG animation and MathML href.

otherdaniel:
- Add SVG <a href> & <a xlink:href> to list of javascript:-attributes.
- Add a list for SVG animations.
- Minor edits when using those lists.

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/212.html" title="Last updated on Jan 10, 2025, 2:32 PM UTC (7134494)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/212/3604c34...otherdaniel:7134494.html" title="Last updated on Jan 10, 2025, 2:32 PM UTC (7134494)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -485,6 +485,29 @@ template contents). It consistes of these steps:
 
 </div>
 
+<div class=note>
+<span class=marker>Note:</span> Current browsers support `javascript:` URLs
+only when navigating. Since navigation itself is not an XSS threat we handle
+navigation to `javascript:` URLs, but not navigations in general.
+
+Declarative navigation falls into a handful of categories:
+
+  1. Anchor elements. (`<a>` in HTML and SVG namespaces)
+  1. Form elements that trigger navigation as part of the form action.
+  1. [[MathML]] allows [any element to act as an anchor](
+     https://www.w3.org/TR/MathML3/mathml.html#chapter2_fund.globatt).
+  1. [[SVG11]] animation.
+
+The first two are covered by the [=built-in navigating URL attributes list=].
+
+The MathML case is covered by a seperate rule, because there is no formalism
+in this spec to cover a "per-namespace global" rule.
+
+The SVG animation case is covered by the
+[=built-in animating URL attributes list=]. But since the interpretation of SVG animation elements depends on the animation target, and since during sanitization we cannot know what the final target will be, the [=sanitize=] algorithm blocks any animation of `href` attributes.
+
+</div>
+
 <div algorithm>
 To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
 1. Let |url| be the result of running the [=basic URL parser=]
--- a/index.bs
+++ b/index.bs
@@ -779,6 +779,21 @@ navigations are "unsafe", are as follows:
     { "`name`" &rightarrow; "`formaction`", "`namespace`" &rightarrow; `null` }
   ],
   <br>
+  [
+    { "`name`" &rightarrow; "`a`", "`namespace`" &rightarrow; [=SVG namespace=] },
+    { "`name`" &rightarrow; "`href`", "`namespace`" &rightarrow; `null` }
+  ],
+  <br>
+  [
+    { "`name`" &rightarrow; "`a`", "`namespace`" &rightarrow; [=SVG namespace=] },
+    { "`name`" &rightarrow; "`href`", "`namespace`" &rightarrow; [=XLink namespace=] }
+  ],
+  <br>
+  [
+    { "`name`" &rightarrow; "`base`", "`namespace`" &rightarrow; [=HTML namespace=] },
+    { "`name`" &rightarrow; "`href`", "`namespace`" &rightarrow; `null` }
+  ],
+  <br>
 ]&raquo;
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -468,9 +468,15 @@ template contents). It consistes of these steps:
          - "data-" is a [=code unit prefix=] of [=Attr/local name=] and
             [=Attr/namespace=] is `null` and
             |configuration|["{{SanitizerConfig/dataAttributes}}"] is true
-      1. If |handleJavascriptNavigationUrls| and &laquo;[|elementName|, |attrName|]&raquo; matches an entry in the
-         [=built-in navigating URL attributes list=], and if |attribute|
-         [=contains a javascript: URL=], then remove |attribute| from |child|.
+      1. If |handleJavascriptNavigationUrls|:
+         1. If &laquo;[|elementName|, |attrName|]&raquo; matches an entry in
+            the [=built-in navigating URL attributes list=], and if |attribute|
+            [=contains a javascript: URL=], then remove |attribute| from |child|.
+         1. If the [=built-in animating URL attributes list=]
+            [=SanitizerConfig/contains=]
+            &laquo;[|elementName|, |attrName|]&raquo; and |attr|'s
+            [=get an attribute value|value=] [=string/is=] "`href`" or
+            "`xlink:href`", then [=/remove an attribute|remove=] |attr|.
 
 </div>
 
@@ -717,11 +723,12 @@ regard to order:
 
 ## Defaults ## {#sanitization-defaults}
 
-There are three builtins:
+There are four builtins:
 
 * The [=built-in safe default configuration=],
 * the [=built-in safe baseline configuration=], and
-* the [=built-in navigating URL attributes list=].
+* the [=built-in navigating URL attributes list=], and
+* the [=built-in animating URL attributes list=].
 
 The <dfn>built-in safe default configuration</dfn> is as follows:
 ```
@@ -797,6 +804,33 @@ navigations are "unsafe", are as follows:
 ]&raquo;
 </div>
 
+The <dfn>built-in animating URL attributes list</dfn>, which can be used in
+[[SVG11]] to declaratively modify navigation elements to use "`javascript:`"
+URLs, is as follows:
+
+&laquo;[
+  <br>
+  [
+    { "`name`" &rightarrow; "`animate`", "`namespace`" &rightarrow; [=SVG namespace=] },
+    { "`name`" &rightarrow; "`attributeName`", "`namespace`" &rightarrow; `null`] }
+  ],
+  <br>
+  [
+    { "`name`" &rightarrow; "`animateMotion`", "`namespace`" &rightarrow; [=SVG namespace=] },
+    { "`name`" &rightarrow; "`attributeName`", "`namespace`" &rightarrow; `null` }
+  ],
+  <br>
+  [
+    { "`name`" &rightarrow; "`animateTransform`", "`namespace`" &rightarrow; [=SVG namespace=] },
+    { "`name`" &rightarrow; "`attributeName`", "`namespace`" &rightarrow; `null` }
+  ],
+  <br>
+  [
+    { "`name`" &rightarrow; "`set`", "`namespace`" &rightarrow; [=SVG namespace=] },
+    { "`name`" &rightarrow; "`attributeName`", "`namespace`" &rightarrow; `null` }
+  ],
+  <br>
+]&raquo;
 
 # Security Considerations # {#security-considerations}
 
--- a/index.bs
+++ b/index.bs
@@ -472,6 +472,11 @@ template contents). It consistes of these steps:
          1. If &laquo;[|elementName|, |attrName|]&raquo; matches an entry in
             the [=built-in navigating URL attributes list=], and if |attribute|
             [=contains a javascript: URL=], then remove |attribute| from |child|.
+         1. If |child|'s [=Element/namespace=] [=string/is=] the
+            [=MathML Namespace=] and |attr|'s [=Attr/local name=] [=string/is=]
+            "`href`" and  |attr|'s [=Attr/namespace=] is `null` or the
+            [=XLink namespace=] and |attr| [=contains a javascript: URL=],
+            then [=/remove an attribute|remove=] |attr|.
          1. If the [=built-in animating URL attributes list=]
             [=SanitizerConfig/contains=]
             &laquo;[|elementName|, |attrName|]&raquo; and |attr|'s
--- a/index.bs
+++ b/index.bs
@@ -431,16 +431,15 @@ template contents). It consistes of these steps:
            parser for which this assertion should hold. If in the future
            this algorithm will be used in different contexts, this assumption
            needs to be re-examined.
-  1. If |child| [=implements=] {{Text}}:
-    1. [=continue=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -305,32 +305,39 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
+      1. If the [=navigating URL attributes list=] [=SanitizerConfig/contains=]
+         &laquo;[|elementName|, |attrName|]&raquo; and "`javascript:`" is a
+         [=code unit prefix=] of the result of [=strip leading and
+         trailing ASCII whitespace=] from |attr|'s [=get an attribute
+         value|value=]:
+         1. Then [=remove an attribute|remove=] |attr|.
+      1. If the [=animating URL attributes list=] [=SanitizerConfig/contains=]
+         &laquo;[|elementName|, |attrName|]&raquo; and |attr|'s
+         [=get an attribute value|value=] [=string/is=] "`href`" or
+         "`xlink:href`":
```
annevk, line 336: We should double check that `xlink:href` is actually meaningful as it supposedly only works when `xmlns` puts `xlink` in scope which is never the case in HTML.

annevk (commented):
@cure53 do you know why DOMPurify bans all SVG animation? Do you know other maintainers we can reach out to?

index.bs
zcorpan: `xlink:href` works in HTML regardless of namespace declarations. https://html.spec.whatwg.org/multipage/parsing.html#adjust-foreign-attributes

securityMB:
> @cure53 do you know why DOMPurify bans all SVG animation? Do you know other maintainers we can reach out to?

I believe the reason is that you can use SVG animation to animate `href` attribute which can lead to XSS.

For example:

```html
<svg>
<animate xlink:href=#x 
         attributeName=href 
         values=javascript:alert(1) />
<a id=x>
  <text x=10 y=10>
    Click me
  </text>
</a>

```

annevk:
@securityMB thanks for weighing in! This PR attempts to tackle that in a more specific way, allowing generic animation features to continue to work. We are wondering if that is good enough or if we'll run into other trouble.

index.bs
annevk: @zcorpan that is `xlink:href` as an attribute, but not as an attribute value of these SVG attributes.

zcorpan: Ah, I see. Still, `xmlns:xlink` is also an adjusted attribute.

annevk: Good point, so it should work when `xmlns:xlink` is also specified. In which case we want to always ban `xlink:href` as is suggested here. (Having tests for all these combinations would be good though.)

I guess the other piece of feedback I forgot to note down here is that we should make sure we have the same processing model as these attributes have today, which regards to splitting on whitespace and such.

lukewarlow:
Trusted types covers the SVGAnimatedString IDL type, which covers you updating an SVGScriptElement hrefs baseVal, and it also includes the `javascript:` url pre-navigation handling for when a javascript URL is defined in markup and navigated to.

So I think if sanitizer is stripping the `javascript:` value for href (and probably it's xlink variant?) from any markup via the safe sync, along with probably stripping out the script element from SVG entirely (aka follow the same model as happens with HTML),  that should be sufficient?

annevk:
@lukewarlow no, see the example above: https://github.com/WICG/sanitizer-api/pull/212#issuecomment-2033918699. We cannot rely on the navigation check Trusted Types has here. We need something along the lines of this PR or ban SVG animation entirely, but that seems less well founded.

lukewarlow:
Sorry to clarify I didn't mean to rely on TT for anything I was just providing context on what TT does for this area.

otherdaniel:
I wrote a test to figure out what browsers will actually do... 

- Animating `"href"` to a `javascript:`-url is implemented by all browsers.
- Chrome & Safari, but not Firefox, will animate `"xlink:href"`, but only if there's an explicit `xmlns:xlink` declaration.
- There seems to be no pre-processing (like strippping whitespace) involved. I couldn't get `"  href  "` to animate.

(Source is [here](https://github.com/web-platform-tests/wpt/pull/45601) and [there](https://wpt.fyi/results/svg/animations/animate-href-with-javascript-url.html?label=pr_head&max-count=1&pr=45601).)

index.bs
```diff
@@ -305,32 +305,46 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
-      1. If &laquo;[|elementName|, |attrName|]&raquo; matches an entry in the
-         [=navigating URL attributes list=], and if |attr|'s [=protocol=] is
-         "`javascript:`":
-         1. Then remove |attr| from |child|.
+         1. Then [=remove an attribute|remove=] |attr|.
+      1. If the [=navigating URL attributes list=] [=SanitizerConfig/contains=]
+         &laquo;[|elementName|, |attrName|]&raquo; and "`javascript:`" is a
+         [=code unit prefix=] of the result of [=strip leading and
+         trailing ASCII whitespace=] from |attr|'s [=get an attribute
+         value|value=]:
```
annevk, line 331: This doesn't work. We need to actually parse the value as a URL (using the basic URL parser to avoid doing anything with `blob:` URLs) and then look at the scheme.

See https://url.spec.whatwg.org/#concept-basic-url-parser.

This has quite a few loopholes:

* "`JAVAscript:`"
* "`java\tscript:`" (replace `\t` with a literal tab)
* "`\0javascript:`" (replace `\0` with U+0000)

etc.

index.bs
```diff
@@ -305,35 +305,74 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
+         1. Then [=remove an attribute|remove=] |attr|.
        1. Call [=sanitize=] on |child|'s [=Element/shadow root=] with |config|.
     1. else:
-      1. [=/remove=] |child|.
+      1. [=/Remove=] |child|.
 
 </div>
 
+<div class=note>
+<span class=marker>Note:</span> Current browsers support `javascript` URLs
```
annevk, line 347: I would make this `javascript:` URLs throughout (including the `:`), including in definitions. I think that's more consistent with how we approach this elsewhere.

index.bs
```diff
@@ -305,35 +305,74 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
+[=animating URL attributes list=]. But since the interpretation of SVG animation elements depends on the animation target, and since during sanitization we cannot know what the final target will be, the [=sanitize=] algorithm blocks any animation of `href` attributes.
+
+</div>
+
+<div algorithm>
+To determine whether an |attribute| <dfn>contains a javascript URL</dfn>, do this:
+
+1. Let |url| be the result of running the [=basic URL parser=]
+   on |attribute|'s [=get an attribute value|value=].
+1. Return whether |url|'s [=url/scheme=] [=string/is=] "`javascript`".
```
annevk, line 371: You need a check here for |url| being failure.

annevk: I think not including the `base` element URL here is probably fine, especially now that the `base` element has its own checks, unless @zcorpan can come up with something contrived?

index.bs
```diff
@@ -273,21 +273,21 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
            this algorithm will be used in different contexts, this assumption
            needs to be re-examined.
   1. If |child| [=implements=] {{Text}}:
-    1. [=continue=].
+    1. Then [=continue=].
```
annevk, line 276: Can we inline these, I think that would be much clearer?

> 1. If child implements Text, then continue.
> 2. Otherwise, if child implements Comment:

(Also, specifications use "Otherwise", not "Else".

index.bs
```diff
@@ -305,35 +305,76 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
+
+</div>
+
+<div algorithm>
+To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>, do this:
+
+1. Let |url| be the result of running the [=basic URL parser=]
+   on |attribute|'s [=get an attribute value|value=].
+1. Let |contains javascript: url| be:
+   |url| is not `failure` and |url|'s [=url/scheme=] [=string/is=] "`javascript`".
```
annevk, line 372: I recommend just returning early if url is failure. I think that would be clearer.

annevk (commented):
I think this looks good now, modulo nits. It's unfortunate we have to invoke the URL parser though, but I guess so be it. We never said that safe was going to be cheap.

index.bs
otherdaniel: Done. (Here and elsewhere.)

index.bs
zcorpan: https://bugzilla.mozilla.org/show_bug.cgi?id=1850967 is not fixed yet, so maybe we can check `base` for that reason.

Longer-term, I think the only attack would be some form of markup injection (`<a ` before `base`), but maybe that's not part of the scenarios we worry about with this API?

index.bs
```diff
@@ -304,36 +303,72 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
+         [=SanitizerConfig/contains=] |attrName|,
+         then [=remove an attribute|remove=] |attr|.
+      1. If the [=navigating URL attributes list=] [=SanitizerConfig/contains=]
+         &laquo;[|elementName|, |attrName|]&raquo; and |attr|
+         [=contains a javascript: URL=],
+         then [=remove an attribute|remove=] |attr|.
+      1. If |child|'s [=Element/namespace=] [=string/is=] the
+         [=MathML Namespace=] and |attr|'s [=Attr/local name=] [=string/is=]
+         "`href`" and  |attr| [=contains a javascript: URL=],
+         then [=remove an attribute|remove=] |attr|.
```
mozfreddyb, line 332: (from our meeting) Maybe `xlink:href` is missing here...

otherdaniel:
In the meeting, we had briefly discussed what the support for MathML + href + javascript: was. It seems there's already WPT tests for this: https://wpt.fyi/results/mathml/relations/html5-tree?label=master&label=experimental&aligned&q=mathml%2Frelations%2Fhtml5-tree%2Fhref-click 

href-click-003.tentative.html tests navigating to a `javascript:`-url.

index.bs
otherdaniel: Added `failure`  handling. Added `<base href>` to the navigating URL attributes list.

index.bs
otherdaniel: Added a check for the namespace (either null or XLink namespace).

otherdaniel:
Since this PR was quasi-abandoned, I've rebased and (partially) re-written it. Also, instead of the usual chronological commit history, this is now seperated out in semantically meaninfgul, seperate commits, that can also be reviewed seperately.

1. A handful of editorial changes that are likely uncontroversial. Mainly, moving if-then clauses with tiny then actions into a single line.
2. Complete hte list of built-in navigating URL attributes.
3. Handle SVG animation of navigating attributes.
4. Handle MathML xlink:href support.
5. Add an explanatory note.

index.bs
```diff
@@ -472,8 +470,54 @@ template contents). It consistes of these steps:
...
+in this spec to cover a "per-namespace global" rule.
+
+The SVG animation case is covered by the
+[=built-in animating URL attributes list=]. But since the interpretation of SVG animation elements depends on the animation target, and since during sanitization we cannot know what the final target will be, the [=sanitize=] algorithm blocks any animation of `href` attributes.
+
+</div>
+
+<div algorithm>
+To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>, do t
+his:
```
mozfreddyb, line 515:
```suggestion
To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>, do
this:
```

mozfreddyb (approved):

index.bs

index.bs
```diff
@@ -472,8 +470,54 @@ template contents). It consistes of these steps:
...
       1. If |handleJavascriptNavigationUrls| and &laquo;[|elementName|, |attrName|]&raquo; matches an entry in the
          [=built-in navigating URL attributes list=], and if |attribute|'s [=protocol=] is
-         "`javascript:`":
-         1. Then remove |attribute| from |child|.
+         "`javascript:`",
+         then [=/remove=] |attribute| from |child|.
+      1. If |handleJavascriptNavigationUrls| and |child|'s [=Element/namespace=]
+         [=string/is=] the [=MathML Namespace=] and |attr|'s
+         [=Attr/local name=] [=string/is=]
+         "`href`" and  |attr|'s [=Attr/namespace=] [=string/is=] `null` or the
```
annevk, line 478: Seems weird to use string/is on null. Just use "is" without reference?

index.bs
```diff
@@ -472,8 +470,54 @@ template contents). It consistes of these steps:
...
+         [=get an attribute value|value=] [=string/is=] "`href`" or
+         "`xlink:href`", then [=/remove an attribute|remove=] |attr|.
+
+</div>
+
+
+<div class=note>
+<span class=marker>Note:</span> Current browsers support `javascript:` URLs
+only when navigating. Since navigation itself is not an XSS threat we treat
+navigation to `javascript:` URLs, but not navigations in general.
```
annevk, line 493: This latter sentence reads a bit weird. Maybe "we handle" instead of "we treat"?

index.bs
```diff
@@ -472,8 +470,54 @@ template contents). It consistes of these steps:
...
+
+</div>
+
+<div algorithm>
+To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>, do
+this:
+
+1. Let |url| be the result of running the [=basic URL parser=]
+   on |attribute|'s [=get an attribute value|value=].
+1. If |url| is `failure` return false.
```
annevk, line 519:
```suggestion
1. If |url| is `failure`, then return false.
```

annevk (commented):
Looks good to me modulo some nits.

index.bs

index.bs
```diff
@@ -431,16 +431,15 @@ template contents). It consistes of these steps:
...
            this algorithm will be used in different contexts, this assumption
            needs to be re-examined.
-  1. If |child| [=implements=] {{Text}}:
-    1. [=continue=].
-  1. else if |child| [=implements=] {{Comment}}:
-    1. If |configuration|["{{SanitizerConfig/comments}}"] is not true:
-      1. [=/remove=] |child|.
-  1. else:
+  1. If |child| [=implements=] {{Text}}, then [=continue=].
+  1. Otherwise if |child| [=implements=] {{Comment}}:
```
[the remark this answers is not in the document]

index.bs

index.bs

annevk (approved):

PR #217, opened 2024-04-09
Use class=extract instead of __TO_BE_MERGED suffix.

### LINKED ISSUE #216: Remove __TO_BE_MERGED variants
`setHTMLUnsafe` and `parseHTMLUnsafe` are in HTML since https://github.com/whatwg/html/pull/9538, with the difference that the optional dictionary isn't there.

What remains to remove these variants from this spec?

I ask because I'm writing a patch to exclude them from `@webref/idl`.

otherdaniel:
In #210, we added a suffix to methods defined in HTML in order to avoid tooling errors. Per #216, this still causes inconvenience. So here we solve this by using `class=extract`, to prevent these partial interfaces from being extracted at all.

See: https://speced.github.io/bikeshed/#no-idl
Fixes #216.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -102,22 +102,18 @@ configuration. The methods come in two by two flavours:
 ## Sanitizer API ## {#sanitizer-api}
 
 The {{Element}} interface defines two methods, {{Element/setHTML()}} and
-{{Element/setHTMLUnsafe__TO_BE_MERGED()}}. Both of these take a {{DOMString}} with HTML
+{{Element/setHTMLUnsafe()}}. Both of these take a {{DOMString}} with HTML
 markup, and an optional configuration.
 
-<pre class=idl>
+<pre class="idl extract">
 partial interface Element {
-  [CEReactions] undefined setHTMLUnsafe__TO_BE_MERGED(HTMLString html, optional SanitizerConfig config = {});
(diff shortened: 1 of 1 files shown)


annevk:
@foolip should prolly review this then.

index.bs
```diff
@@ -102,22 +102,18 @@ configuration. The methods come in two by two flavours:
 ## Sanitizer API ## {#sanitizer-api}
 
 The {{Element}} interface defines two methods, {{Element/setHTML()}} and
-{{Element/setHTMLUnsafe__TO_BE_MERGED()}}. Both of these take a {{DOMString}} with HTML
+{{Element/setHTMLUnsafe()}}. Both of these take a {{DOMString}} with HTML
 markup, and an optional configuration.
 
-<pre class=idl>
+<pre class="idl extract">
```
foolip, line 108: This will also prevent `setHTML` from being extracted. Is that OK, should they go together?

annevk: Yeah, that seems fine.

otherdaniel: My thinking was that a "partial" integration wouldn't help us much and that we'd simply take out all the bits where we extend interfaces defined in HTML.

foolip: OK, that seems OK. LGTM!

PR #218, opened 2024-04-15
Fix `Exposed` statement in dfn of `Sanitizer` interface

tidoust:
Identifiers of an extended attribute need to be enclosed in parentheses in Web IDL: https://webidl.spec.whatwg.org/#dfn-extended-attribute


otherdaniel:
Thank you! Didn't know that.. :)

Issue #233, opened 2024-06-12
Solving safe/unsafe defaults


annevk:
We have a problem currently that `new Sanitizer()` ends up being the same configuration for safe and unsafe, with completely different effects. This makes it really hard to design an API to manipulate the policy as the implications for safe and unsafe are different.

To solve this I think we could do the following:

1. Introduce `Sanitizer.default()` that returns the configuration we want `setHTML()` to use when no configuration is provided.
2. Make `setHTML()`'s `SetHTMLOptions`'s `sanitizer` default to "`default`" rather than `{}`. `setHTMLUnsafe()` will continue to default to `{}`. "`default`" here is an enum value that means `Sanitizer.default()` and is needed to make things work IDL-wise. It will also provide us with a way forward for other named policies we might want to add in the future.
3. `new Sanitizer()` now represents an empty policy. So if you pass that to `setHTML()` you only get the XSS-blocklist impacting you.
4. We can continue to keep `get()` and `getUnsafe()`. `get()` is useful to see how your policy will be impacted by the XSS-blocklist.

This means that if you do `setHTML("...", sanitizer: {})` you're only impacted by the XSS-blocklist. If you do `setHTML("...")` you get our more conservative "`default`" take.

Coupled with #229 I think this would provide all the building blocks we need.

annevk:
3/4 are addressed by https://github.com/WICG/sanitizer-api/commit/7e2f1271e46e645a3ae00b86d79d05934a29c4c5 and 1 is noted. But 1/2 still need to be done.

And note that 4 was morphed into the clearly better `removeUnsafe()`.

otherdaniel:
We discussed this in the meeting and decided to not add a method. I've closed the PR. Closing this issue, too, as there's nothing left to do.

Issue #229, opened 2024-05-15
extend() static and instance method for Sanitizer


annevk:
To address the problem in #228 I suggest we add an `extend()` static (and instance) method that takes a configuration as argument. However, in this configuration the allow and blocklists serve as additions and subtractions to the existing configuration. (So you can have both `elements` and `blockElements` for instance.)

The `extend()` static method is a convenience for constructing a Sanitizer object without any argument passed (giving you the default) and then invoking the instance `extend()` method upon that with the argument passed.

otherdaniel:
In the last meeting, we briefly discussed which options we might have here. Basic designs I can think of:

1. "config arithmetic": Take multiple configs, and provide methods to intersect or join them.
1. "builders": Have manipulation methods on the config object.
1. try to "lower" the config representation to match existing platform primitives, so that the config description itself becomes easy to modify.

------------

1. "Config Arithmetic"

   The Sanitizer gains two methods, `Sanitizer.extend` and `Sanitizer.restrict`, that create a new sanitizer (or maybe modify `this`) that is the union (`Sanitizer.extend`) or the intersection ( `Sanitizer.restrict`) of the two inputs, with the (informal) definition that any node in the parse result would be allowed if any input (in case of union) or all inputs (in case of intersection) allows it.
  
   One nice property here is that the "ensure safety" operation for the safe methods could be reduced to an implicit `.restrict` call against a built-in allow-list.

1. "Builders"

   The Sanitizer gains methods that loosely corresponds to the keys in the config, `Sanitizer.allow`, `Sanitizer.remove`, `Sanitizer.replaceWithChildren`, which take a parameter describing one element / attribute, as string or dictionary, just like the config does. These update the sanitizer to allow / remove / replaceWithChildren the respective element.

1. "Lowering"

   The current implementation does a lot of checks - e.g. that no name is being used twice - that makes it unnecessarily difficult to modify an existing config.

   E.g., to add an allowed element, one can retrieve a config from an existing Sanitizer as a dictionary and then add that element name to the allow list. But: Because the Sanitizer config will throw on duplicate use of a name, one has to first check whether that given element is already mentioned in any of the other config items; and remove it if so; and only then can one add it.

   Maybe one could (maybe optionally?) remove those checks, in order to make it easy to manipulate configs with platform methods.

otherdaniel:
My current favourite is adding restrict/extend to the config object:

[...]

Issue #228, opened 2024-05-15
Safe sanitizer default


annevk:
We already have a difference between what is allowed by default and what is truly blocked. A `foo` element is not allowed by default, but you could allow it as it's not an XSS risk. I.e., there's a backing XSS blocklist that makes that distinction possible.

We discussed adding `iframe` to this backing XSS blocklist a week ago, but what if we kept it a pure XSS blocklist, and instead blocked a lot more by default to make it truly safe:

* Styling-based attacks
* Automatic fetching of resources
* DOM clobbering: id/name attributes
* Form submission
* `<base>`
* Nested documents
* `<meta>`

This needs some more thought:

* Links. I think we should allow single-click attacks, but maybe we need to look carefully at `target` and opener? Probably the best is to allow `<a href>` and not `<a target rel>`. Most other attributes are probably okay?

The one tricky aspect is that this even further exposes a weakness in the current setup that it's "hard" to manipulate the default configuration. You have to extract it as per https://wicg.github.io/sanitizer-api/#configobject and then manipulate it. Which is a little annoying.

As discussed in the meeting we might need to have a better way to manipulate configurations from the get go.

benbucksch:
You're thinking in the right direction. The use cases will be generally almost the same, but often slightly different in the details, so allowlist and whitelist (as discussed in #229) in addition to a pre-config is the right solution.

You're mentioning a few interesting cases. I concur with others that protection against XSS attacks should be the baseline and an opt-out should be hard or come with warnings.

On top of this minimal baseline, there are many additional requirements, depending on the use cases. Many of them cannot be meaningfully expressed by allowing or blocking specific tags or attributes. "Automatic fetching of resources" is a very good example. Compare https://github.com/cure53/DOMPurify/issues/951 and the reasons why it's not practically possible to use blocklists for that. Additionally, in an HTML email application like Thunderbird, the user wants a switch that allows external resources only for specific emails, so a sanitizer feature like this should be easy to switch on or off.

I would suggest to
a) have multiple pre-configs to choose from (e.g. including or excluding SVG and MathML)
b) optional on/off switches on top of it (e.g. load external resources, nested docs etc.)
c) allow and block lists on top of it

benatkin:
Some things to look out for can be found in https://github.com/w3c/webappsec/issues/656 such as `<link rel="prefetch">`.

annevk:
This done through "`default`".

PR #237, opened 2024-08-13
Modify-able config.

otherdaniel:
[This isn't quite ready yet. Using a PR for tool support and discussion.]

This re-writes the config logic, based on discussions in the recent meetings, and with the intent to address issues #228, #229, and #233.

The corner stones are:

- The Sanitizer (and thus its config) become mutable. It has methods to add stuff to  any of the allow or deny lists.
- The actual config processing is now defined in terms of those methods.
- There is only one "safe-ify" operation, and it gets called once (for safe methods). This is the only different between safe and unsafe versions.
  - (Well, this is only a half-truth. The handling of javascript:-navigation URLs remains an odd-ball that is magically different between "safe" and "unsafe" versions.)
- The configs have an "other" setting, which allows them to express "everything" and "nothing" filters, which in turn allows us to treat the unsafe defaults as just another regular case.
- The entire "known" logic is gone. But in turn, the "baseline" is now back, and is a deny-list.

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/237.html" title="Last updated on Nov 28, 2024, 5:42 PM UTC (e05db5a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/237/85d25b5...otherdaniel:e05db5a.html" title="Last updated on Nov 28, 2024, 5:42 PM UTC (e05db5a)">Diff</a>

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -378,11 +378,11 @@ The following methods are offered on the Sanitizer object:
   - `options` is an optional dictionary argument.
     Supported keys are: `"attributes":` and `"removeAttributes":.`
 - `removeElement(x)`
-- `replaceWithChildren(x)`
+- `replaceElementWithChildren(x)`
 - `allowAttribute(x)`
 - `removeAttribute(x)`
-- `comments(bool)`
-- `dataAttributes(bool)`
+- `setComments(bool)`
+- `setDataAttributes(bool)`
 
 These correspond 1:1 to the keys in the configuration dictionary.
 
--- a/explainer.md
+++ b/explainer.md
@@ -381,6 +381,8 @@ The following methods are offered on the Sanitizer object:
 - `replaceWithChildren(x)`
 - `allowAttribute(x)`
 - `removeAttribute(x)`
+- `comments(bool)`
+- `dataAttributes(bool)`
 
 These correspond 1:1 to the keys in the configuration dictionary.
 
@@ -400,6 +402,23 @@ s.get();  // { elements: ["div", "p", "span"], removeElements: ["b"] }
           // Really, all these entries will be dictionaries with name and
           // namespace entries.
 ```
+
+If one wishes to modify the element-dependent attributes, then `allow` is
+the way to do this, with a dictionary as argument. This allows `"attributes"`
+and `"removeAttributes"` keys, like the configuration dictionary. These
+element-dependent attributes are set, meaning they overwrite any previously
+set values, rather than some sort of merger operation.
+
+```js
+const s = new Sanitizer();
+s.element({name: "div", attributes: ["id", "class"]});
+s.element({name: "div", attributes: ["style"]});
+// s now allows <div style="bla">, but will drop the id= from <div id="bla">
+```
+
+Since the configuration is mutable, passing around a pre-configured Sanitizer
+can be used to let other callers modify its configuration. The "safe" methods
+(`setHTML` and `parseHTML`) will still guarantee XSS safety.
 
 ### Configuration Errors
 
--- a/explainer.md
+++ b/explainer.md
@@ -374,9 +374,10 @@ functionality of the Sanitizer.
 
 The following methods are offered on the Sanitizer object:
 
-- `allow(x, options)`
-  - `options` is an optional dictionary argument.
-    Supported keys are: `"attributes":` and `"removeAttributes":.`
+- `allowElement(x)`
+  - `x` can be a dictionary (similar to all other methods), but it also
+    supports additional keys to allow (`"attributes"`) or to remove attributes
+    ("removeAttributes"`) for this particular element type.
 - `removeElement(x)`
 - `replaceElementWithChildren(x)`
 - `allowAttribute(x)`
@@ -396,7 +397,7 @@ namespace, just as with the configuration dictionary.
 
 ```js
 const s = new Sanitizer({ elements: ["div", "p", "b"] });
-s.element("span");
+s.allowElement("span");
 s.removeElement("b");
 s.get();  // { elements: ["div", "p", "span"], removeElements: ["b"] }
           // Really, all these entries will be dictionaries with name and
@@ -411,8 +412,8 @@ set values, rather than some sort of merger operation.
 
 ```js
 const s = new Sanitizer();
-s.element({name: "div", attributes: ["id", "class"]});
-s.element({name: "div", attributes: ["style"]});
+s.allowElement({name: "div", attributes: ["id", "class"]});
+s.allowElement({name: "div", attributes: ["style"]});
 // s now allows <div style="bla">, but will drop the id= from <div id="bla">
 ```
 
--- a/explainer.md
+++ b/explainer.md
@@ -363,6 +363,44 @@ element.setHTML("XXX<!-- Hello world! -->XXX", {sanitizer: config_comments});
 // <div>XXX<!-- Hello world! -->XXX</div>
 ```
 
+### Modifying Existing Configurations
+
+The `Sanitizer` object offers multiple methods to easily modify or tailor
+an existing configuration. The query methods (`get()` and `getUnsafe()`) can
+be used to retrieve a dictionary representation of a Sanitizer,
+for introspection, or for use with the Sanitizer constructor to create a new
+Sanitizer. Additionally, there are methods that directly manipulate the filter
+functionality of the Sanitizer.
+
+The following methods are offered on the Sanitizer object:
+
+- `allow(x, options)`
+  - `options` is an optional dictionary argument.
+    Supported keys are: `"attributes":` and `"removeAttributes":.`
+- `removeElement(x)`
+- `replaceWithChildren(x)`
+- `allowAttribute(x)`
+- `removeAttribute(x)`
+
+These correspond 1:1 to the keys in the configuration dictionary.
+
+Adding an element or attribute to any of the allow- or deny-lists will also
+remove that element or attribute from the other lists for its type. E.g.,
+calling `allow(x)` will also remove `x` from the removeElements and
+replaceWithChildrenElements lists.
+
+Any name can be given as either a string, or a dictionary with name or
+namespace, just as with the configuration dictionary.
+
+```js
+const s = new Sanitizer({ elements: ["div", "p", "b"] });
+s.element("span");
+s.removeElement("b");
+s.get();  // { elements: ["div", "p", "span"], removeElements: ["b"] }
+          // Really, all these entries will be dictionaries with name and
+          // namespace entries.
+```
+
 ### Configuration Errors
 
 The configuration allows expressing redundant or even contradictory options.
--- a/index.bs
+++ b/index.bs
@@ -25,7 +25,6 @@ spec:html; type:dfn; text: template contents
 </pre>
 <pre class="anchors">
 text: window.toStaticHTML(); type: method; url: https://msdn.microsoft.com/en-us/library/cc848922(v=vs.85).aspx
-text: internal slot; type:dfn; url: https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots
 text: parse HTML from a string; type: dfn; url: https://html.spec.whatwg.org/#parse-html-from-a-string
 </pre>
 <pre class="biblio">
@@ -228,8 +227,8 @@ dictionary SetHTMLOptions {
 </pre>
 
 The {{Sanitizer}} configuration object encapsulates a filter configuration.
-The same config can be used with both safe or unsafe methods, where the safe
-methods perform an implicit {{removeUnsafe}} operation. The intent is
+The same configuration can be used with both safe or unsafe methods, where
+the safe methods perform an implicit {{removeUnsafe}} operation. The intent is
 that one (or a few) configurations will be built-up early on in a page's
 lifetime, and can then be used whenever needed. This allows implementations
 to pre-process configurations.
@@ -240,7 +239,7 @@ It can also be modified directly.
 <pre class=idl>
 [Exposed=(Window,Worker)]
 interface Sanitizer {
-  constructor(optional SanitizerConfig config = {});
+  constructor(optional SanitizerConfig configuration = {});
 
   // Query configuration:
   SanitizerConfig get();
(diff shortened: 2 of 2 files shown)


mozfreddyb:
We discussed this in a synchronous meetings. I'm copying the notes verbatim so we can continue async.

> 
> - Document.parseHTML / Document.parseHTMLUnsafe - How does it handle safety?
> - We should not need `otherMarkup`
> - get/getUnsafe should not be a method
> - Some method for turning unsafe into safe
> - Don’t want to call it “safe”, because we don’t know what safe is
> - Eg. “WithBaseline” (baseline is applied to config)
> - Element vs allowAttribute?
> - setComment why set-method instead of an attribute-setter?
> - Should we use allow instead of allowElement (or element)
> - Anne: How to undo replaceWithChildren
> - Tom: element() currently removes from the replaceWithChildren list
> - Guillaume: “Other name for something captured by allow or block list”
> - We have a baseline know
> - Annevk: baseline remove all unsafe stuff
> - Default “safe list” of allowed elements
> - Annevk: Expose the default list with a static method
> - Anne: Name the method removeUnsafe to make a config safe (instead of safeify, because we can’t easily guarantee safeness for all use cases)
> - It shouldn't be replaceWithChildren but replaceWithChildrenElements (as per earlier group consensus)
> - Anne: Rename methods to addElement, addRemoveElement, addReplaceWithChildrenElement, etc. (prefix with add to make it clear there is a modification of the instance)
> - No group concensus for the `add` prefix. We need to discuss this further.
> - Tom: Make setComment a property like comment
> - Anne: This would imply introspection for the properties and not for other like elements, removeElements etc.
> - Anne: Remove getUnsafe, instead a method that makes sanitizer safe (removeUnsafe)
> - Guillaume: get() returns whatever the current state is
> - Anne: Could we use toJSON instead of get?
> - Guillaume/Frederik: We need to worry about having functions that are not JSON compatible in the future (e.g. if we add a fallback feature in v2)
> - all: Let’s keep get then.
>

mozfreddyb (changes requested):
Really great progress so far. We have some nits (and some larger questions) in the review notes, please see above.

otherdaniel:
- Added issue links to issues for naming and other markup
- Added "unsafe" handling to parseHTMLUnsafe.
- Replaced `getUnsafe` query method with `removeUnsafe` modifier.
- Naming mostly deferred. But made sure all method names start with a verb.
- Finished handling of element-dependent attributes.
- Various editorial / link-ification improvements.

index.bs
```diff
@@ -217,47 +228,157 @@ dictionary SetHTMLOptions {
...
+
+ISSUE(240): "other markup" TBD.
+
+ISSUE: Should these be setter methods -- particularly the setXXX(boolean) --
+    or setters or properties or somesuch?
+
+ISSUE: Should the modifier methods return a reference to [=this=], so that you
+    can 'chain' methods?
+    (e.g. `sanitizer.allowElement("a").allowElement("span")`).
+
```
mozfreddyb, line 272: Please file issues for those that are only in the file

index.bs
```diff
@@ -217,47 +228,157 @@ dictionary SetHTMLOptions {
...
 <div algorithm>
 The <dfn for="Sanitizer" export>constructor</dfn>(|config|)
 method steps are:
 
-1. Store |config| in [=this=]'s [=internal slot=].
+1. Let |valid| be the return value of [=set a config|Set=] |config| on [=this=].
+1. If not |valid|, throw a {{TypeError}1. If not |valid|, throw a {{TypeError}}}
```
mozfreddyb, line 278:
```suggestion
1. If not |valid|, throw a {{TypeError}.
```

index.bs
```diff
@@ -297,6 +418,7 @@ dictionary SanitizerConfig {
 };
 </pre>
 
+ISSUE: Sould members be required, or have declared defaults?
```
mozfreddyb, line 421: (Same as above, pls file ISSUE for those that area only in the document)

index.bs
```diff
@@ -308,40 +430,53 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
 
 </div>
 
 ## Sanitization Algorithms ## {#sanitization}
 
-<div algorithm="sanitize">
+<div algorithm>
 For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
-[=SanitizerConfig/canonical=] {{SanitizerConfig}} |config|, run these steps:
+{{Sanitizer}} |sanitizer| and a [=boolean=] |safe|, run these steps:
```
annevk, line 464:
```suggestion
{{Sanitizer}} |sanitizer|, and a [=boolean=] |safe|, run these steps:
```

index.bs
```diff
@@ -308,40 +430,53 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
+
+1. Let |config| be the value of |sanitizer|'s [=internal slot=].
+1. If |safe|, let |config| be the result of calling [=remove unsafe=] on |config|.
```
annevk, line 467:
```suggestion
1. If |safe| is true, then set |config| to the result of calling [=remove unsafe=] on |config|.
```

index.bs
```diff
@@ -358,324 +493,120 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
+
+To <dfn for="SanitizerConfig">remove unsafe</dfn> from a |config|, do this:
+
+1. [=Assert=]: The [=built-in safe baseline config=] has
+   {{SanitizerConfig/removeElements}} and {{SanitizerConfig/removeAttributes}}
+   keys set, but not {{SanitizerConfig/elements}},
+   {{SanitizerConfig/replaceWithChildrenElements}}, or
+   {{SanitizerConfig/attributes}}.
+1. Let |result| be a copy of |config|.
+1. [=list/For each=] |elem| in
```
annevk, line 552: Sometimes you use elem, sometimes, item, and sometimes element. I would settle on element for all of these. I'd generally prefer less abbreviations in prose, so also attribute and configuration instead of attr and config.

index.bs
```diff
@@ -358,324 +493,120 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
+   [=built-in safe baseline config=][{{SanitizerConfig/removeElements}}]:
+    1. Call |result|.{{Sanitizer/removeElement()|removeElement}}(|elem|)
```
annevk, line 554: You cannot call public API from this algorithm. You instead need to define an internal algorithm that removeElement() and this can both call.

index.bs
```diff
@@ -358,324 +493,120 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
-1. Let |result| be a new [=ordered set=].
-2. [=list/iterate|For each=] |name| in |list|, call
-   [=canonicalize a sanitizer name=] on |name| with |defaultNamespace| and
-   [=set/append=] to |result|.
-3. Return |result|.
+To <dfn for="Sanitizer">set a config</dfn> |config| on a {{Sanitizer}} |sanitizer|, do this:
+
+1. [=Assert=]: |config| is a [=dictionary=].
+1. [=list/iterate|For each=] |item| of |config|[{{SanitizerConfig/elements}}] do:
+  1. Call |sanitizer|.{{Sanitizer/allowElement()|allowElement}}(|item|).
```
annevk, line 567: Same problem as above. You cannot call public API here.

index.bs

index.bs
otherdaniel: Done. (Also removed issues that had been resolved, like this one, or otherMarkup.

index.bs
```diff
@@ -217,47 +228,157 @@ dictionary SetHTMLOptions {
...
-1. Return the result of [=canonicalize a configuration=] with the value of
-   [=this=]'s [=internal slot=] and true.
+1. Return the value of [=this=]'s [=internal slot=].
+
+</div>
+
+<div algorithm>
+The <dfn for="Sanitizer" export>allowElement</dfn>(|element|) method steps are:
+
+1. Let |name| be the result of [=canonicalize a sanitizer name=] |element| with [=HTML namespace=] as the default namespace.
```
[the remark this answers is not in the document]
otherdaniel: Hmm. Canonicalize a sanitizer name is used for both elements and attributes. I could add a "canonicalize an elemnt" and a "canonicale an attribute", both of which would then canonicalize their name, but it seems like that merely adds an indirection. (At least as no other canonicalization steps are performed.)

[Leaving open for now.]

index.bs
```diff
@@ -217,47 +228,157 @@ dictionary SetHTMLOptions {
...
+1. [=list/Remove=] |name| from [=this=]'s [=internal slot=]'s {{SanitizerConfig/elements}}.
```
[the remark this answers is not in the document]
otherdaniel: Added a [=SanitizerConfig/remove=], to explain removal of a name structure.

Removed "internal slot" in favour of configuration; and systematically added dictionary syntax to reference its members.

(I think I cargo-culted the "internal slot" from ECMAScript, which defines an "internal slot" as a hook that the embedder can use to attach data to a script object. I agree that doesn't actually make any sense over here.)

index.bs

index.bs
otherdaniel: Mostly done. I consistently use element, attribute, and configuration now. I still use item now and then, but only for contexts where it could be either an element or an attribute. Happy to rename that, but I couldn't really come up with a nice name for element_or_attribute.

index.bs

mozfreddyb:
copying over from #238 

change
-   replaceWithChildrenElements to replaceElementWithChildren()
-   setComments() to plural.
 

keep:
-  removeElement()
-   allowElement()
-   removeAttribute()
-   allowAttribute()
-   setDataAttributes()

otherdaniel:
Updated the PR with feedback from today's meeting:

- Adopted the updated naming.
- Introduced canonicalization methods for element with attributes; elements; and attributes. Rewrote the caller algorithms so that they canonicalize in the first step, and then only use the canonicalized value.

index.bs
```diff
@@ -217,48 +227,86 @@ dictionary SetHTMLOptions {
 </pre>
 
 The {{Sanitizer}} configuration object encapsulates a filter configuration.
-The same config can be used with both safe or unsafe methods. The intent is
+The same configuration can be used with both safe or unsafe methods, where
```
annevk, line 230:
```suggestion
The same configuration can be used with both safe and unsafe methods, where
```
Perhaps we should add quotation marks around "safe"?

index.bs
```diff
@@ -217,48 +227,86 @@ dictionary SetHTMLOptions {
...
+the safe methods perform an implicit {{removeUnsafe}} operation. The intent is
```
annevk, line 231:
```suggestion
the safe methods perform an implicit {{removeUnsafe}} operation on the passed in configuration and have a default configuration when none is passed. The intent is
```

index.bs
```diff
@@ -217,48 +227,86 @@ dictionary SetHTMLOptions {
...
-to build a new configuration based on an existing one.
+The configuration object can be queried to return a configuration dictionary.
+It can also be modified directly.
 
 <pre class=idl>
 [Exposed=(Window,Worker)]
 interface Sanitizer {
-  constructor(optional SanitizerConfig config = {});
+  constructor(optional SanitizerConfig configuration = {});
+
```
annevk, line 243: Let's add a TODO here for `[NewObject] static Sanitizer getDefault()`? (I suppose it could also return a `SanitizerConfig`, but an instance seems better.)

index.bs
```diff
@@ -217,48 +227,86 @@ dictionary SetHTMLOptions {
...
+The <dfn for="Sanitizer" export>allowElement</dfn>(|element|) method steps are to [=allow an element=] with |element| and [=this=]'s [=Sanitizer/configuration=].
+</div>
+
+<div algorithm>
+The <dfn for="Sanitizer" export>removeElement</dfn>(|element|) method steps are
+to [=remove an element=] with |element| and [=this=]'s [=Sanitizer/configuration=].
+</div>
 
+<div algorithm>
+The <dfn for="Sanitizer" export>replaceElementWithChildren</dfn>(|element|) method steps are to [=replace an element=] with |element| and [=this=]'s [=Sanitizer/configuration=].
```
annevk, line 286: Let's also call the internal operation "replace an element with children"?

index.bs
```diff
@@ -308,40 +355,53 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
+1. [=Assert=]: |options|["`sanitizer`"] is either a {{Sanitizer}} instance
    or a [=dictionary=].
 1. If |options|["`sanitizer`"] is a {{Sanitizer}} instance:
-   1. Then let |config| be the value of |options|["`sanitizer`"]'s [=internal slot=].
-   1. Otherwise let |config| be the value of |options|["`sanitizer`"].
-1. Return the result of calling [=canonicalize a configuration=] on
-   |config| and |safe|.
+   Then return  |options|["`sanitizer`"].
+1. [=Assert=]: |options|["`sanitizer`"] is a [=dictionary=].
+1. Return new {{Sanitizer}}(|options|["`sanitizer`"]).
```
annevk, line 381: Public API.

annevk (approved):
At this point I would be okay with landing this to get the public draft in a better state, but we'll need follow-up issues for `setHTML()` defaulting and such (and any comments in this PR that you decide to postpone addressing).

index.bs
mozfreddyb: Bikeshed. Sorry: "unsafe and non-unsafe is terrible". How about saying safe and linking to Security Considerations that starts with a bit more of an explicit "This is what we consider safe for our scenario."?

annevk: I'm not sure anyone suggested unsafe and non-unsafe?

mozfreddyb: I suggested & non-suggested that. Nevermind that bit.

I saw your criticism and agree with you that just saying safe without quotes or any extra context could be ambiguous. Putting it in quotes is not a lot of more context either

Considering that, I suggest putting it in quotes and "safe" linking the Security Considerations section would make more sense to me.

index.bs
```diff
@@ -308,40 +361,58 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
-1. Assert: |options| is a [=dictionary=].
-1. If |options|["`sanitizer`"] doesn't [=map/exist=], then return undefined.
-1. Assert: |options|["`sanitizer`"] is either a {{Sanitizer}} instance
+To <dfn for="SanitizerConfig">get a sanitizer instance from options</dfn> for
+an options dictionary |options|, do:
+
+1. [=Assert=]: |options| is a [=dictionary=].
+1. If |options|["`sanitizer`"] doesn't [=map/exist=], then:
+   1. Let |result| be a new {{Sanitizer}} instance.
+   1. Call [=set a config=] with an empty [=dictionary=] on |result|.
```
annevk, line 382: "set a configuration"

index.bs
```diff
@@ -353,330 +424,229 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
 1. Return |result|.
 
 </div>
 
 <div algorithm>
-In order to <dfn>canonicalize a sanitizer element list</dfn> |list|, with a
-default namespace |defaultNamespace|, run the following steps:
+To <dfn for="Sanitizer">set a config</dfn> |configuration| on a {{Sanitizer}} |sanitizer|, do this:
+
+1. [=Assert=]: |configuration| is a [=dictionary=].
```
annevk, line 576: Instead of asserting this here, you should specify the argument type above.

> To **set a config** given a dictionary |configuration| and a Sanitizer |sanitizer|:

explainer.md
```diff
@@ -363,6 +363,63 @@ element.setHTML("XXX<!-- Hello world! -->XXX", {sanitizer: config_comments});
...
+be used to retrieve a dictionary representation of a Sanitizer,
+for introspection, or for use with the Sanitizer constructor to create a new
+Sanitizer. Additionally, there are methods that directly manipulate the filter
+functionality of the Sanitizer.
+
+The following methods are offered on the Sanitizer object:
+
+- `allow(x, options)`
+  - `options` is an optional dictionary argument.
+    Supported keys are: `"attributes":` and `"removeAttributes":.`
```
annevk, line 379:
```suggestion
- `allowElement(x, options)`
```

index.bs
otherdaniel: Done. (Technically, a note instead todo, since that's apparently what bikeshed knows about.)

index.bs
otherdaniel: Added a definition to "safe and unsafe" where we first use the terms, and linked there. Added scare quotes, too.

index.bs

index.bs
otherdaniel: Done. Added a pronoun, too, "with its children", since this is just plain text and I think that makes it a little clearer what the operation actually does.

index.bs

index.bs
```diff
@@ -308,40 +355,53 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
-an options dictionary |options| and a boolean |safe|, do:
+To <dfn for="SanitizerConfig">get a sanitizer instance from options</dfn> for
+an options dictionary |options|, do:
 
-1. Assert: |options| is a [=dictionary=].
-1. If |options|["`sanitizer`"] doesn't [=map/exist=], then return undefined.
-1. Assert: |options|["`sanitizer`"] is either a {{Sanitizer}} instance
+1. [=Assert=]: |options| is a [=dictionary=].
+1. If |options|["`sanitizer`"] doesn't [=map/exist=],
+   then return new {{Sanitizer}}().
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -308,40 +355,53 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
-<div algorithm="sanitize">
+<div algorithm>
 For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
-[=SanitizerConfig/canonical=] {{SanitizerConfig}} |config|, run these steps:
+{{Sanitizer}} |sanitizer|, and a [=boolean=] |safe|, run these steps:
+
+1. Let |configuration| be the value of |sanitizer|'s [=Sanitizer/configuration=].
+1. If |safe| is true, then set |configuration| to the result of calling [=remove unsafe=] on |configuration|.
+1. Call [=sanitize core=] on |node|, |configuration|, and |safe| (as value for
+    handling javascript navigation urls).
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -308,40 +355,53 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
+
+</div>
+
+<div algorithm>
+The <dfn>sanitize core</dfn> operation,
+using a {{ParentNode}} |node|, a {{SanitizerConfig}} |configuration|, and a
+[=boolean=] |handle javascript navigation urls|, iterates over the DOM tree
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -353,330 +413,229 @@ For the main <dfn>sanitize</dfn> operation, using a {{ParentNode}} |node|, a
...
+To <dfn for="Sanitizer">remove an attribute</dfn> |attribute| from a {{SanitizerConfig}} |configuration|, do:
+
+1. Set |attribute| to the result of [=canonicalize a sanitizer attribute=] with |attribute|.
+1. [=SanitizerConfig/Add=] |attribute| to |configuration|["{{SanitizerConfig/removeAttributes}}"].
+1. [=SanitizerConfig/Remove=] |attribute| from |configuration|["{{SanitizerConfig/attributes}}"].
+
+</div>
+
+<div algorithm>
+To <dfn for="Sanitizer">allow comments</dfn> with |allow| on a {{SanitizerConfig}} |configuration|, do:
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -725,71 +687,75 @@ Equality for [=ordered sets=] is equality of its members, but without
...
+* the [=built-in unsafe default config=],
+* the [=built-in safe baseline config=], and
+* the [=navigating URL attributes list=].
+
+The <dfn>built-in safe default config</dfn> is the same as the [=built-in safe baseline config=].
+
+ISSUE(233): Determine if this actually holds.
+
+
+The <dfn>built-in unsafe default config</dfn> is meant to allow anything.
```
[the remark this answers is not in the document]
otherdaniel: I suspect not. I left this one, since the next step is to re-work the default handling along the lines we discussed in yesterday's meeting, and I hope I can fix all the builtin/default things in one go thhere.

index.bs
```diff
@@ -725,71 +687,75 @@ Equality for [=ordered sets=] is equality of its members, but without
...
-  { name: "class", namespace: null },
-  ...
-]
+{
+  removeElements: [
+    { name: "script", namespace: "http://www.w3.org/1999/xhtml" },
+    { name: "script", namespace: "http://www.w3.org/2000/svg" }
+  ],
+  removeAttributes: [....],
+  comments: true,
```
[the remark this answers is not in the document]

index.bs

index.bs
otherdaniel: Done. That indeed reads nicer.

index.bs
```diff
@@ -308,40 +361,58 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
-1. If |options|["`sanitizer`"] doesn't [=map/exist=], then return undefined.
-1. Assert: |options|["`sanitizer`"] is either a {{Sanitizer}} instance
+To <dfn for="SanitizerConfig">get a sanitizer instance from options</dfn> for
+an options dictionary |options|, do:
+
+1. [=Assert=]: |options| is a [=dictionary=].
+1. If |options|["`sanitizer`"] doesn't [=map/exist=], then:
+   1. Let |result| be a new {{Sanitizer}} instance.
+   1. Call [=set a config=] with an empty [=dictionary=] on |result|.
+   1. [=Assert=]: The return value is true.
```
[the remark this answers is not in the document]

explainer.md
otherdaniel: Done. (Also fixed a few more cases in the same chapter.)

index.bs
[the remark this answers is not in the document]
otherdaniel: Done now. In all cases, the algorithms update the argument with the canonicalized version and use only that.

index.bs

otherdaniel:
I think I resolved all the issues. Will land tomorrow, if no more comments come in tonight.

Next step will be to rework the default handling, and to put up some doc and/or poposal about the built-in lists.

mozfreddyb:
🎈

PR #243, opened 2024-11-29
Update the default handling

otherdaniel:
SetHTMLOptions's "sanitizer" key can now also contain an enum SanitizerPresets, which has only one member, "default". Which, it turns out, is the default value. This is now the only way to invoke the defaults. An empty dictionary would then
be a Sanitizer which allows anything. (Minus the "removeUnsafe" content, when used with a "safe" method.)

This follows discussion at the 2024-11-27 meeting.

The explainer still needs updaing.

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/243.html" title="Last updated on Dec 11, 2024, 4:03 PM UTC (f267f67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/243/7e2f127...otherdaniel:f267f67.html" title="Last updated on Dec 11, 2024, 4:03 PM UTC (f267f67)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -385,12 +385,12 @@ Note: This algorithm works for both {{SetHTMLOptions}} and
 
 1. Let |sanitizerSpec| be "{{SanitizerPresets/default}}".
 1. If |options|["{{SetHTMLOptions/sanitizer}}"] [=map/exists=], then:
-   1. Let |sanitizerSpec| be |options|["{{SetHTMLOptions/sanitizer}}"]
+   1. Set |sanitizerSpec| to |options|["{{SetHTMLOptions/sanitizer}}"]
 1. [=Assert=]: |sanitizerSpec| is either a {{Sanitizer}} instance,
    a [=string=] which is a {{SanitizerPresets}} member, or a [=dictionary=].
 1. If |sanitizerSpec| is a [=string=]:
    1. [=Assert=]: |sanitizerSpec| [=is=] "{{SanitizerPresets/default}}"
-   1. Let |sanitizerSpec| be the [=built-in safe default configuration=].
+   1. Set |sanitizerSpec| to the [=built-in safe default configuration=].
 1. [=Assert=]: |sanitizerSpec| is either a {{Sanitizer}} instance,
    or a [=dictionary=].
 1. If |sanitizerSpec| is a [=dictionary=]:
@@ -398,7 +398,7 @@ Note: This algorithm works for both {{SetHTMLOptions}} and
    1. Let |setConfigurationResult| be the result of [=set a configuration=]
       with |sanitizerSpec| on |sanitizer|.
    1. If |setConfigurationResult| is false, [=throw=] a {{TypeError}}.
-   1. Let |sanitizerSpec| be |sanitizer|.
+   1. Set |sanitizerSpec| to |sanitizer|.
 1. [=Assert=]: |sanitizerSpec| is a {{Sanitizer}} instance.
 1. Return |sanitizerSpec|.
 
--- a/index.bs
+++ b/index.bs
@@ -259,11 +259,11 @@ interface Sanitizer {
 
   // Remove markup that executes script. May modify multiple lists:
   undefined removeUnsafe();
+
+  // Create a Sanitizer instance using the built-in defaults.
+  [NewObject] static Sanitizer getDefault();
 };
 </pre>
-
-Note: {{Sanitizer}} will likely get an additional method:
-    <br>`[NewObject] static Sanitizer getDefault();`
 
 A {{Sanitizer}} has an associated <dfn for="Sanitizer">configuration</dfn>, a {{SanitizerConfig}}.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -222,8 +222,9 @@ The family of {{Element/setHTML()}}-like methods all accept an options
 dictionary. Right now, only one member of this dictionary is defined:
 
 <pre class=idl>
+enum SanitizerPresets { "default" };
 dictionary SetHTMLOptions {
-  (Sanitizer or SanitizerConfig) sanitizer = {};
+  (Sanitizer or SanitizerConfig or SanitizerPresets) sanitizer = "default";
```
annevk, line 227: This does not work as it would make the behavior the same for `setHTMLUnsafe()` as `setHTML()`. I think we only want to give the latter SanitizerPresets for now and the default.

This also makes me think of a slightly different API shape. What if the `Sanitizer` constructor took `SanitizerPresets`? Then we would not need `Sanitizer.getDefault()` and possibly other static methods in the future. The only problem is that "`default`" would no longer be a good name as the default `Sanitizer` is `{}`.

index.bs
```diff
@@ -315,6 +316,11 @@ update [=this=]'s [=Sanitizer/configuration=] with the result of calling [=remov
 on [=this=]'s [=Sanitizer/configuration=].
 </div>
 
+<div algorithm>
+The <dfn for="Sanitizer" export>getDefault</dfn> method steps are to return the
```
annevk, line 320:
```suggestion
The static <dfn for="Sanitizer" export>getDefault</dfn> method steps are to return the
```

index.bs
otherdaniel: > This does not work as it would make the behavior the same for `setHTMLUnsafe()` as `setHTML()`. I think we only want to give the latter SanitizerPresets for now and the default.

Oh, I overlooked that. Is there a way to do that without copying SetHTMLOptions, or will I need two of those? (`SetHTMLOptions`/`SetHTMLUnsafeOptions`).
 
> This also makes me think of a slightly different API shape. What if the `Sanitizer` constructor took `SanitizerPresets`? Then we would not need `Sanitizer.getDefault()` and possibly other static methods in the future. The only problem is that "`default`" would no longer be a good name as the default `Sanitizer` is `{}`.

I guess `(SanitizerConfig or SanitizerPresets)`? I think that makes sense.

otherdaniel: Uploaded a new draft, with SetHTMLOptions and SetHTMLUnsafeOptions. Also changed the constructor to take a config or presets.

index.bs
```diff
@@ -242,7 +246,7 @@ It can also be modified directly.
 <pre class=idl>
 [Exposed=(Window,Worker)]
 interface Sanitizer {
-  constructor(optional SanitizerConfig configuration = {});
+  constructor(optional (SanitizerConfig or SanitizerPresets) configuration = "default");
```
annevk, line 249: I was initially thinking this would be bad because we would want the Sanitizer to be empty, but I actually rather like this. And you can still get an empty sanitizer easily, by using `new Sanitizer({ })`.

index.bs
```diff
@@ -261,15 +265,15 @@ interface Sanitizer {
...
-
 A {{Sanitizer}} has an associated <dfn for="Sanitizer">configuration</dfn>, a {{SanitizerConfig}}.
 
 <div algorithm>
 The <dfn for="Sanitizer" export>constructor</dfn>(|configuration|)
 method steps are:
 
+1. If |configuration| is a {{SanitizerPresets}} [=string=], then:
+    1. [=Assert=]: |configuration| [=is=] {{SanitizerPresets/default}}.
+    1. Let |configuration| be the [=built-in safe default configuration=].
```
annevk, line 276:
```suggestion
    1. Set |configuration| to the [=built-in safe default configuration=].
```

index.bs
```diff
@@ -315,6 +319,11 @@ update [=this=]'s [=Sanitizer/configuration=] with the result of calling [=remov
 on [=this=]'s [=Sanitizer/configuration=].
 </div>
 
+<div algorithm>
+The <dfn for="Sanitizer" export>getDefault</dfn> method steps are to return the
+result of [=get a sanitizer instance from options=] with &laquo;[ "{{SetHTMLOptions/sanitizer}}" &rightarrow; "{{SanitizerPresets/default}}" ]&raquo;
+</div>
+
```
annevk, line 326:
```suggestion
```

index.bs
otherdaniel: I'm using "Let |...| be" quite a lot. Not sure where I had picked this up. Is that generally undesired; or only here (where I'm updating a variable, rather than initializing it)?

index.bs
```diff
@@ -373,25 +377,30 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
...
-      with an empty [=dictionary=] on |result|.
-   1. [=Assert=]: The |setConfigurationResult| is true.
-   1. Return |result|.
-1. [=Assert=]: |options|["`sanitizer`"] is either a {{Sanitizer}} instance
+Note: This algorithm works for both {{SetHTMLOptions}} and
+    {{SetHTMLUnsafeOptions}}. They only differ in the defaults.
+
+1. Let |sanitizerSpec| be "{{SanitizerPresets/default}}".
+1. If |options|["{{SetHTMLOptions/sanitizer}}"] [=map/exists=], then:
+   1. Let |sanitizerSpec| be |options|["{{SetHTMLOptions/sanitizer}}"]
```
annevk, line 388:
```suggestion
   1. Set |sanitizerSpec| to |options|["{{SetHTMLOptions/sanitizer}}"]
```
(already used let for this)

annevk: This happens a few times in this algorithm it seems.

annevk (approved):
Left some additional nits, but generally this looks great to me. Thanks!

index.bs

PR #244, opened 2024-11-29
Sanitizer built-ins document

otherdaniel:
This is meant as a starting point for our built-ins. It's a "mostly free-form" document, in that it's just text with one line per element, plus markdown-style headings. The idea is to classify elements and attributes into groups.

I started out putting everything into "other", and then moving them into better defined groups. The idea is to work down the "other" list until it's empty.

I copied all the elements from the "proposed allow lists" doc into the "harmless" category. Will do the same for attributes.

Source(s):
-  Elements & attributes lists from Chrome. (Which may include legacy elements no longer supported.)
- "Proposed allow lists -- Sanitizer API - 2024-03" document prepared by Frederik.

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/244.html" title="Last updated on Jan 13, 2025, 4:10 PM UTC (4f94295)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/244/9254522...otherdaniel:4f94295.html" title="Last updated on Jan 13, 2025, 4:10 PM UTC (4f94295)">Diff</a>


mozfreddyb:
FYI, this PR is still called "Not ready yet". Let me know when you are seeking review.

otherdaniel:
> FYI, this PR is still called "Not ready yet". Let me know when you are seeking review.

About now would be good, so I renamed it. :) 

I added more "sections" with spec links, also for attributes. I think that should cover all of HTML; while SVG + MathML coverage is still a bit of a mess. There are a lot more leftover ("other") attributes than there were with elements.

Changes so far are:
- I copied the per-element attributes from the HTML spec, for any element that I expect to be default-allowed. (It's a manual process, so I was trying to save myself some time.)
- I also copied global HTML attributes + aria attributes from the spec(s), including spec links.
- I removed all attributes that are used locally somewhere from the global list. I'm not sure this is quite correct.
- All of this was manual, so I wouldn't be surprised if there are some omissions somewhere. I'm unsure how to do QA here.

annevk (commented):
Thanks for making this list. Very useful to have something based on the HTML standard. Here's an initial very conservative safelist for "`default`":

- html
- head
- title
- All of Sections (body, article, ...)
- All of Grouping Content (p, hr, ...)
- Most of Text-level Semantics:
   - a attributes target, referrerpolicy, download, and ping I would omit
- All of Edits (ins, del)
- All of Tabular Data
- SVG & MathML: TBD
- Global attributes:
   - dir
   - lang
   - title

mozfreddyb:
👍 to what @annevk says, that we should build upon the HTML spec rather than casting the widest net.

As written in #245, I could see us being a bit more iterative by using his relatively small list for now and discussing additions individually as they come up (which they are bound to anyway).

otherdaniel:
- Updated list according to Anne's suggestion.
- Changed the list format a little, and added script that turns it into JSON.
- Included this and baseline config from the spec.
- Moved the builtin files (json, text, script) to a builtins/ directory.

builtins/safe-baseline-configuration.json
```diff
@@ -0,0 +1,147 @@
...
+    "onvalidationstatuschange",
+    "onvolumechange",
+    "onwaiting",
+    "onwebkitanimationend",
+    "onwebkitanimationiteration",
+    "onwebkitanimationstart",
+    "onwebkitfullscreenchange",
+    "onwebkitfullscreenerror",
+    "onwebkittransitionend",
+    "onwheel"
```
mozfreddyb, line 145: What should we do here? In spec purity terms, I believe we should stick to [those in the HTML standard](https://html.spec.whatwg.org/#:~:text=List%20of%20event%20handler%20content%20attributes) and make a big note that many engines support non-standardized and add them as a hint or such?
But In reality, I can see this going wrong.

@evilpie: How would we best identify the list of supported event handler attributes in Gecko?

evilpie: In Gecko, Trusted Types currently [uses](https://searchfox.org/mozilla-central/rev/6f4bc07bd9eb07ed34774841ecf3b1f3678791e8/dom/security/trusted-types/TrustedTypeUtils.cpp#540-549) us the [EventNameList.h](https://raw.githubusercontent.com/mozilla/gecko-dev/f009b3332db1c5197b1eb9c7108f68f9bb0900a8/dom/events/EventNameList.h)

builtins/safe-baseline-configuration.json
```diff
@@ -0,0 +1,147 @@
+{
+  "removeElements": [
+    {
+      "namespace": "http://www.w3.org/1999/xhtml",
+      "name": "script"
+    },
+    {
+      "namespace": "http://www.w3.org/2000/svg",
+      "name": "script"
+    }
```
mozfreddyb, line 10: We need to include frame/iframe/object/embed/svg+use

otherdaniel: Done.

builtins/safe-baseline-configuration.json
annevk: I think iterating over them is okay. We might have to revisit this when upstreaming.

index.bs
```diff
@@ -718,24 +721,36 @@ There are three builtins:
...
+
+<div class="note" role="note">
+
+<span class="marker">Note:</span> The [=remove unsafe=] algorithm specifies
+to additionally remove any [=event handler content attributes=], as defined
+in [[HTML]].
+If a [=user agent=] defines extensions to the [[HTML]] spec with additional
+[=event handler content attributes=], it is its responsibility to decide how
+to handle them. Using the current [=event handler content attributes=] list,
+the safe baseline configuration looks effectively like so:
```
mozfreddyb, line 746: Nice, this is very important.
nit picking, but is there a higher level of severity than a "note"? Can we do warnings? :D

mozfreddyb (approved):

index.bs
otherdaniel: > Nice, this is very important. nit picking, but is there a higher level of severity than a "note"? Can we do warnings? :D

The tool offers [Notes, Issues, Examples, Advisements](https://speced.github.io/bikeshed/#notes-etc)

HTML suports [highly dramatic warnings](https://html.spec.whatwg.org/#typographic-conventions), with a few usages in the spec, e.g. [here](https://html.spec.whatwg.org/#content-type-sniffing).

If we want to use a warning here, I'd just emulate the CSS in the spec directly.

mozfreddyb: Guess it's best to do what's easiest and not worth nit picking too hard. :) thank you!

annevk: You could add a source comment to upgrade this to `class=warning` when upstreaming.

otherdaniel: There's also a pre-defined CSS style for "annoying-warning", but that sounds.. like too much. :)

mozfreddyb:
lgtm 👍

index.bs
mozfreddyb: I'm happy with this 🥳  thank you!

annevk (approved):
This seems fine. I'm a bit surprised we block nested documents in baseline though. Did we discuss that and I forgot about it?

otherdaniel:
> This seems fine. I'm a bit surprised we block nested documents in baseline though. Did we discuss that and I forgot about it?

We discussed it at end of last year; see minutes of the 2024-12-11 call. At the time, there was a rather large "other" category that we went through. The minutes record "some gut calls during the call", and "no" for frame-related stuff.  I do remember this as gut calls indeed, with noone having super strong opinions one way or another.

annevk:
Thanks! Admittedly it's also somewhat hard to do a nested document policy that makes sense, so once there's demand for that we can try to figure out more dedicated syntax I suppose. Or point people towards "unsafe".

otherdaniel:
Landing this. I'll note that the list format makes this really easy to change, spec-wise. :)

Issue #249, opened 2025-01-10
Spec bugs / clarifications


otherdaniel:
Collection of issues I encountered when cross-checking our (early-stage) implementation & test results against the spec.

These might well be bugs in the test suite or impl, rather than the spec, though. I'm just listing things here so we won't forget...

otherdaniel:
Config procesing: We currently throw when an element or global attribute is listed in both allow & remove, but not for per-element attributes.

```js
new Sanitizer({elements: ["p"], removeElements: ["p"]})  // throws
new Sanitizer({elements: [{name: "p", attributes: ["id"], removeAttributes: ["id"]}]})  // does not throw.
```

I think that's just an oversight.

otherdaniel:
"comments" and "dataAttribute" keys in the config default to `false`. That's because WebIDL defaults to `false` when accessing a missing boolean member. This could be handled differently by checking for presence.

I thought we had decided this differently; but not sure I remember correctly.

```js
bla.setHTML("<p>hello <!-- world -->", {})  // Comment is dropped: <p>hello.
```

mozfreddyb:
> Config procesing: We currently throw when an element or global attribute is listed in both allow & remove, but not for per-element attributes.
> 
> ```js
> new Sanitizer({elements: ["p"], removeElements: ["p"]})  // throws
> new Sanitizer({elements: [{name: "p", attributes: ["id"], removeAttributes: ["id"]}]})  // does not throw.
> ```
> 
> I think that's just an oversight.

I could see someone arguing that the second case could work for clear semantics on which of the properties is first being checked, leading to either  `p` with `id` in place, but all other elements without any `id` **or*  `p` element stays but all `id` attributes removed.

Not sure what to prefer, to be honest.



> "comments" and "dataAttribute" keys in the config default to `false`. That's because WebIDL defaults to `false` when accessing a missing boolean member. This could be handled differently by checking for presence.
> 
> I thought we had decided this differently; but not sure I remember correctly.
> 
> ```js
> bla.setHTML("<p>hello ", {})  // Comment is dropped: <p>hello.
> ```

Wait, what is `{}` doing? I believe we argued the following
[...]

PR #254, opened 2025-01-22
Defaults for boolean SanitizerConfig items

otherdaniel:
Give explicit defaults to SanitizerConfig.comments and SanitizerConfig.dataAttributes.

This fixes one of the issues in #249.

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/254.html" title="Last updated on Jan 31, 2025, 10:35 AM UTC (15cf1fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/254/8e8c366...otherdaniel:15cf1fc.html" title="Last updated on Jan 31, 2025, 10:35 AM UTC (15cf1fc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -200,7 +200,7 @@ The <dfn for="Document" export>parseHTMLUnsafe</dfn>(|html|, |options|) method s
 1. Set |document|'s [=allow declarative shadow roots=] to true.
 1. [=Parse HTML from a string=] given |document| and |compliantHTML|.
 1. Let |sanitizer| be the result of calling [=get a sanitizer instance from options=]
-   with |options|.
+   with |options| and false.
 1. Call [=sanitize=] on |document|'s [=tree/root|root node=] with |sanitizer| and false.
 1. Return |document|.
 
@@ -216,7 +216,7 @@ The <dfn for="Document" export>parseHTML</dfn>(|html|, |options|) method steps a
 1. Set |document|'s [=allow declarative shadow roots=] to true.
 1. [=Parse HTML from a string=] given |document| and |html|.
 1. Let |sanitizer| be the result of calling [=get a sanitizer instance from options=]
-   with |options|.
+   with |options| and true.
 1. Call [=sanitize=] on |document|'s [=tree/root|root node=] with |sanitizer| and true.
 1. Return |document|.
 
@@ -356,8 +356,8 @@ dictionary SanitizerConfig {
   sequence&lt;SanitizerAttribute> attributes;
   sequence&lt;SanitizerAttribute> removeAttributes;
 
-  boolean comments = true;
-  boolean dataAttributes = true;
+  boolean comments;
+  boolean dataAttributes;
 };
 </pre>
 
@@ -372,7 +372,7 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
    |contextElement|'s [=Element/namespace=] is the [=HTML namespace=] or the
    [=SVG namespace=], then return.
 1. Let |sanitizer| be the result of calling [=get a sanitizer instance from options=]
-   with |options|.
+   with |options| and |safe|.
 1. Let |newChildren| be the result of the HTML [=fragment parsing algorithm steps=]
    given |contextElement|, |html|, and true.
 1. Let |fragment| be a new {{DocumentFragment}} whose [=node document=] is |contextElement|'s [=node document=].
@@ -384,7 +384,7 @@ To <dfn>set and filter HTML</dfn>, given an {{Element}} or {{DocumentFragment}}
 
 <div algorithm>
 To <dfn for="SanitizerConfig">get a sanitizer instance from options</dfn> from
-a [=dictionary=] |options|, do:
+a [=dictionary=] |options| with a [=boolean=] |safe|, do:
 
 Note: This algorithm works for both {{SetHTMLOptions}} and
     {{SetHTMLUnsafeOptions}}. They only differ in the defaults.
@@ -402,7 +402,7 @@ Note: This algorithm works for both {{SetHTMLOptions}} and
 1. If |sanitizerSpec| is a [=dictionary=]:
    1. Let |sanitizer| be a new {{Sanitizer}} instance.
    1. Let |setConfigurationResult| be the result of [=set a configuration=]
-      with |sanitizerSpec| on |sanitizer|.
+      with |sanitizerSpec| and |safe| on |sanitizer|.
    1. If |setConfigurationResult| is false, [=throw=] a {{TypeError}}.
    1. Set |sanitizerSpec| to |sanitizer|.
 1. [=Assert=]: |sanitizerSpec| is a {{Sanitizer}} instance.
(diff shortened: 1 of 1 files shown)


annevk (commented):
I don't like this.

1. If boolean fields have a default it should be false, not true.
2. This is strictly less safe.

I think keeping them optional and using that third state to pick a default depending on which method you chose is still a reasonable way forward.

If everyone really wants this behavior we'd have to name them `removeComments` and `removeDataAttributes` I think, defaulting to false.

mozfreddyb:
I was also under the impression that we would make comments & data attributes allowable but not enabled in the `setHTML()` default. 

I think we only discussed a webidl-semantics question on our call on Wednesday, so me skipping this may have contributed to that confusion?

I think we wanted {} to be an "empty sanitizer" but the default to be "a securely configured sanitizer". Is that part of this change?

otherdaniel:
I seem to recall a quite explicit discussion about a `true` default in the meeting. But anyways, the observable behaviours are: (Result in the comments is what we have currently specified, as far as i understand.)

```js
/* 1*/ new Sanitizer().get().comments;  // false
/* 2*/ new Sanitizer({}).get().comments;  // false
/* 3*/ new Sanitizer({comments: true}).get().comments;  // true
/* 4*/ new Sanitizer({comments: false}).get().comments;  // false

/* 5*/ div.setHTMLUnsafe("/*bla*/");  // <div></div>
/* 6*/ div.setHTMLUnsafe("/*bla*/",{sanitizer: {}});  // <div></div>
/* 7*/ div.setHTMLUnsafe("/*bla*/",{sanitizer:{comments:true}});  // <div>/*bla*/</div>
/* 8*/ div.setHTMLUnsafe("/*bla*/",{sanitizer:{comments:false}});  // <div></div>
 
/* 9*/ div.setHTML("/*bla*/");  // <div></div>
/*10*/ div.setHTML("/*bla*/",{sanitizer: {}});  // <div></div>
/*11*/ div.setHTML("/*bla*/",{sanitizer:{comments:true}});  // <div>/*bla*/</div>
/*12*/ div.setHTML("/*bla*/",{sanitizer:{comments:false}});  // <div></div>
```

I take it:
  * 5 is problematic, as it contradicts existing HTML-spec'ed behaviour of `setHTMLUnsafe.`
  * 5 + 6 are expected to be consistent, because `setHTMLUnsafe` defaults to `{}`.
  * 9 can be different from 10, due to the new `"default"` mechanism, but I take it 1 + 2 + 5 + 6 + 10 are meant to be consistent.
  * I take it safe/unsafe, i.e. 6-8 and 10-12, are expected to be consistent except for "baseline" behaviour (non-allowable + javascript:-URLs). Comments aren't a "baseline" thing, so I'd expect them to be fully consistent regarding comments.
 
I'm trying to reconcile these.

I'm guessing the explicit ones (3+4, 7+8, 11+12) are all uncontroversial. The rub is that `setHTMLUnsafe` already has behaviour. I think that behaviour allows comments. And then we have various requirements of methods to be consistent. In particular, I thought we wanted `setHTML` and `setHTMLUnsafe` to be consistent except for clearly articulated differences.

Maybe I'm overlooking something. I think there's other ways to fix this, too, e.g. not defaulting to `{}` for `setHTMLUnsafe`.

annevk:
Nit: you want `<!-- bla -->` instead of `/* bla */`.

I think your analysis is correct. I tend to think it's okay for `Sanitizer`, `setHTML()`, `setHTMLUnsafe()` to handle the dictionary members `comments` and `dataAttributes` differently, but only when they are not explicitly set. I do see that creates some inelegance in the overall model, but overall that seems preferable over not removing them in `setHTML()`.

As such I think `Sanitizer` and `setHTMLUnsafe()` should default them to false, when they are not set. And `setHTML()` should default them to true.

This gives these results:

1. false
2. false
3. true
4. false
5. false
6. false
7. true
8. false
9. true
10. true
11. true
12. false

(This can only happen when you pass a `SanitizerConfig`. Once you have a `Sanitizer` they're always set of course (and they're no longer dictionary members).)

otherdaniel:
Okay, that makes sense.

>  I do see that creates some inelegance in the overall model, but overall that seems preferable over not removing them in `setHTML()`. [...]  And `setHTML()` should default them to true.

This seems contradictory, but I think it's just about notation. I take `comments: true` as allowing comments, so I read the first bit as default-removing them in setHTML (aka, `comments: false`) and  second as default-allowing them in setHTML (aka, `comments: true`). 

> Nit: you want `<!-- bla -->` instead of `/* bla */`.

Oh, yes. I'll fix that... :)

annevk:
Oh yes. I keep inverting them. Hopefully that's just a me problem.

mozfreddyb:
Ha, good catch. I was also thinking "comments" meaning "allow comments". Interesting that even ourselves can get confused 😬

otherdaniel:
I uploaded a new PR that follows the discussion here. The default now depends on safe/unsafe usage.

index.bs
```diff
@@ -641,8 +641,16 @@ To <dfn for="Sanitizer">set a configuration</dfn>, given a [=dictionary=] |confi
   1. Call [=allow an attribute=] with |attribute| and |sanitizer|.
 1. [=list/iterate|For each=] |attribute| of |configuration|["{{SanitizerConfig/removeAttributes}}"] do:
   1. Call [=Sanitizer/remove an attribute=] with |attribute| and |sanitizer|.
-1. Call [=set comments=] with |configuration|["{{SanitizerConfig/comments}}"] and |sanitizer|.
-1. Call [=set data attributes=] with |configuration|["{{SanitizerConfig/dataAttributes}}"] and |sanitizer|.
+1. If |configuration|["{{SanitizerConfig/comments}}"] [=map/exists=]:
+  1. Then call [=set comments=] with |configuration|["{{SanitizerConfig/comments}}"]
+      and |sanitizer|
```
annevk, line 646:
```suggestion
      and |sanitizer|.
```

annevk (approved):
Thanks!

(This PR made me think that maybe we should improve Infra so it's easier to retrieve a map member value and default to some other value if the member doesn't exist.)

mozfreddyb:
✅ 😉

otherdaniel:
What does this do:
```js
div.setHTMLUnsafe("<div data-foo='bar'>", {sanitizer:{dataAttributes:false}});
```

I had expected `<div></div>`, but when testing it against our in-progress implementation it returned `<div data-foor="bar"></div>`. The reason is that an empty `attributes` allow (and empty `removeAttributes`) list allows any attribute, and so the `dataAttributes:` key essentially becomes a no-op since and data attributes are already allows. I think this is indeed what we have specced. Is this what we want? It feels odd to explicitly dis-allow data attributes, but then have them survive anyhow. :)

evilpie:
Okay, maybe this doesn't make sense, but if we made `comments' nullable `boolean'`, wouldn't that more accurately represent the three states of explicit true or false and "implicit based on safe"? If this doesn't make sense from a WebIDL perspective, just ignore me.

bkardell:
>  I was also thinking "comments" meaning "allow comments". Interesting that even ourselves can get confused

Fwiw, I _also_ thought that. Should the name be clearer? I like short names, but I slightly wonder if we are talking about something critical like - maybe its worth some extra characters? `stripComments` or whatever?

otherdaniel:
> Okay, maybe this doesn't make sense, but if we made `comments` a nullable `boolean?`, wouldn't that more accurately represent the three states of explicit true or false and "implicit based on safe"? If this doesn't make sense from a WebIDL perspective, just ignore me.

I think the dictionary already provides this explicit nullability, by not having the key present at all. I.e., `{comments: true}`, `{comments: false}`, or `{}`.

An observable difference would be something like, `{sanitizer: {comments: undefined}}`. I think WebIDL would convert that to `{comments: false}` (i.e. a fixed result), while a nullable boolean would preserve the null-ness (i.e., result depending on safe). That said, I find WebIDL difficult to grok, so maybe I'm misreading that.

Issue #262, opened 2025-03-03
Should we normalize the empty string namespace to null?


evilpie:
I believe as the specification is written currently, we treat the empty string and `null` as two distinct namespaces.
i.e. `new Sanitizer({ elements: [{ name: "p", namespace: null }] })` vs `new Sanitizer({ elements: [{ name: "p", namespace: "" }] })`

Is that something we should change?

@otherdaniel points out that both of the following invocations produce `null`:
```javascript
document.createAttributeNS(null, "foo").namespaceURI  // null
document.createAttributeNS("", "foo").namespaceURI // null
```

It might not be possible to even have an empty namespace URI? The following statement seems to contradict this however for XML:

>  The empty string, though it is a legal URI reference, cannot be used as a namespace name. 

https://www.w3.org/TR/xml-names/#iri-use

annevk:
It probably makes the most sense to follow what the DOM standard does. Accept `DOMString?` and replace the empty string with null as an early normalization step. It's not a great design, but it is what it is.

otherdaniel:
Precedence is here: https://dom.spec.whatwg.org/#validate-and-extract
I don't think we can call that directly (because we don't want to support prefixes here), but we can certainly adopt that first step.

annevk:
Yeah agreed. Perhaps at some point we should abstract that step as "normalize a namespace", but copying it for now seems sufficient.

PR #265, opened 2025-03-05
Normalize the empty string as namespace to be the null namespace.

otherdaniel:
This follows discussion in #262 and precedent in
https://dom.spec.whatwg.org/#validate-and-extract

Fixes: #262

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -707,7 +707,7 @@ return the result of [=canonicalize a sanitizer name=] with |element| and the [=
 
 <div algorithm>
 In order to <dfn>canonicalize a sanitizer attribute</dfn> a
-{{SanitizerAttribute}} |attribute|, 
+{{SanitizerAttribute}} |attribute|,
 return the result of [=canonicalize a sanitizer name=] with |attribute| and `null` as the default namespace.
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -718,9 +718,11 @@ namespace |defaultNamespace|, run the following steps:
 1. [=Assert=]: |name| is either a {{DOMString}} or a [=dictionary=].
 1. If |name| is a {{DOMString}}, then return &laquo;[ "`name`" &rightarrow; |name|, "`namespace`" &rightarrow; |defaultNamespace|]&raquo;.
 1. [=Assert=]: |name| is a [=dictionary=] and |name|["name"] [=map/exists=].
+1. Let |namespace| be |name|["namespace"] if it [=map/exists=], otherwise |defaultNamespace|.
+1. If |namespace| is `""`, set it to null.
```
annevk, line 722:
```suggestion
1. If |namespace| is the empty string, then set it to null.
```

index.bs
```diff
@@ -707,7 +707,7 @@ return the result of [=canonicalize a sanitizer name=] with |element| and the [=
 
 <div algorithm>
 In order to <dfn>canonicalize a sanitizer attribute</dfn> a
-{{SanitizerAttribute}} |attribute|, 
+{{SanitizerAttribute}} |attribute|,
 return the result of [=canonicalize a sanitizer name=] with |attribute| and `null` as the default namespace.
```
annevk, line 711:
```suggestion
return the result of [=canonicalize a sanitizer name=] with |attribute| and null as the default namespace.
```

Issue #264, opened 2025-03-05
Attribute "dir" is duplicated in the "built-in safe default configuration"


evilpie:
We end up with a duplicate `dir` attribute in the [built-in safe default configuration](https://wicg.github.io/sanitizer-api/#built-in-safe-default-configuration) "attributes" list. This is caused by having both a [MathML](https://github.com/WICG/sanitizer-api/blob/33bfc9d308a7e659c16844a4ae5b4127187aaa85/builtins/safe-default-configuration.txt#L239) and [HTML](https://github.com/WICG/sanitizer-api/blob/33bfc9d308a7e659c16844a4ae5b4127187aaa85/builtins/safe-default-configuration.txt#L167) global `dir` attribute in `safe-default-configuration.txt`. (These attributes are both in the null namespace, so they are completely indistinguishable in context of the Sanitizer API)

otherdaniel:
Agreed. It might be easiest to just check for duplicates in the script that generates it. That way, we can keep the source document readable.

otherdaniel:
I forgot about #252 and that I need to manually re-run the build scripts in order to see their effect. Will add another PR. :/

mozfreddyb:
thank you

PR #266, opened 2025-03-05
Check for dupes when generating the default config.

otherdaniel:
The `dir` attribute is in both MathML and HTML globals. The current script will just add it twice. This adds a check for an equal item that already exists.

Fixes: #264


annevk (approved):
Nit: it might be good to log the duplicates somewhere so if they change over time we can be sure we like the implications.

PR #269, opened 2025-03-06
Actually make the sanitize core algorithm recursive

evilpie:
It seems like we never actually recurse into the children for normal elements.

This does duplicate work for `replaceWithChildrenElements`, but I don't think that should worry us from a specification point of view.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -434,8 +434,7 @@ using a {{ParentNode}} |node|, a {{SanitizerConfig}} |configuration|, and a
 beginning with |node|, and may recurse to handle some special cases (e.g.
 template contents). It consistes of these steps:
 
-1. Let |current| be |node|.
-1. [=list/iterate|For each=] |child| in |current|'s [=tree/children=]:
+1. [=list/iterate|For each=] |child| in |node|'s [=tree/children=]:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -434,8 +434,7 @@ using a {{ParentNode}} |node|, a {{SanitizerConfig}} |configuration|, and a
 beginning with |node|, and may recurse to handle some special cases (e.g.
 template contents). It consistes of these steps:
 
-1. Let |current| be |node|.
-1. [=list/iterate|For each=] |child| in |current|'s [=tree/children=]:
+1. [=list/iterate|For each=] |child| in |node|'s [=tree/children=]:
```
annevk, line 437:
```suggestion
1. [=list/iterate|For each=] |child| of |node|'s [=tree/children=]:
```

annevk (commented):
It seems this still gets called even if an earlier step already removed child.

Which is probably also not observable, but I think we should add some early returns for clarity as a specification shouldn't be overly confusing or needlessly inefficient either.

evilpie:
> Which is probably also not observable, but I think we should add some early returns for clarity as a specification shouldn't be overly confusing or needlessly inefficient either.

Maybe in a new PR?

annevk:
I guess that's reasonable, but then we should have an issue tracking it.

annevk (approved):

PR #275, opened 2025-03-19
Immediately continue in sanitize core after removing or replacing a node

evilpie:
Fixes #274

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/275.html" title="Last updated on Mar 19, 2025, 11:25 AM UTC (5e2c250)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/275/445eca6...evilpie:5e2c250.html" title="Last updated on Mar 19, 2025, 11:25 AM UTC (5e2c250)">Diff</a>

PR #277, opened 2025-03-19
Handle replaceWithChildrenElements before the elements allowlist in sanitize core

### LINKED ISSUE #261: Allow replaceWithChildrenElements together with elements
For a config like this: `{elements: ["em"], replaceWithChildrenElements: ["div"]}` and the following HTML `<div><em>Apple<em></div>`, I would expect the result to be `<em>Apple</em>`. 

However, currently [*sanitize-core*](https://wicg.github.io/sanitizer-api/#sanitize-core) first removes all elements not contained in `"elements"` (Step 2.4.2.)  and then handles `"replaceWithChildrenElements"` (Step 2.4.3.). I propose we swap these two steps.

(We already disallow having the same element in both elements and replaceWithChildrenElements)

evilpie:
Fixes #261

Currently, using a replaceWithChildrenElements list together with an elements list does not really work.
Because we first remove all elements that are not in the elements list, which means there is nothing to replace,
because it's impossible for an element to be in both lists.

Example:
```
setHTML("<p><i>hi<i></p>") { elements: ["i"], replaceWithChildrenElements: ["p"] })
```

Before this would just remove the `<p>` element. After this change they are replaced with their allowed child elements
(i.e. `<i>`).

Depends on #275

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/277.html" title="Last updated on Mar 19, 2025, 3:53 PM UTC (49b26b0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/277/8265187...evilpie:49b26b0.html" title="Last updated on Mar 19, 2025, 3:53 PM UTC (49b26b0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -448,14 +448,14 @@ template contents). It consistes of these steps:
   1. Otherwise:
     1. Let |elementName| be a {{SanitizerElementNamespace}} with |child|'s
        [=Element/local name=] and [=Element/namespace=].
-    1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=SanitizerConfig/contains=] |elementName|, or if |configuration|["{{SanitizerConfig/elements}}"] is not [=list/empty=] and does not [=SanitizerConfig/contain=] |elementName|:
-       1. [=/Remove=] |child|.
-       1. [=Continue=].
     1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=SanitizerConfig/contains=] |elementName|:
       1. Call [=sanitize core=] on |child| with |configuration| and
           |handleJavascriptNavigationUrls|.
       1. Call [=replace all=] with |child|'s [=tree/children=] within |child|.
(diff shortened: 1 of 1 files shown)


mozfreddyb (approved):
good catch 😅

mozfreddyb:
Given this is a relatively simple change that moves the order of steps in the  algorithm and because we have a clear test case that won't work without this change, I'm going ahead and merging right away.

Can you add a test for this to WPT, @evilpie?

evilpie:
> Can you add a test for this to WPT, @evilpie?

Actually some current WPT already require this behavior :smiley: 

https://searchfox.org/mozilla-central/rev/4ce36232b265b53de4fb7eb754430f94e262bbbe/testing/web-platform/tests/sanitizer-api/sethtml-tree-construction.sub.dat#134-143

Issue #271, opened 2025-03-10
Duplication of global attributes as element specific attributes in default config


evilpie:
In the default config we are duplicating the global (allowed) attributes: the `title` attribute in `abbr` and `dfn` and the `dir` attribute in `bdi` and `bdo`.

annevk:
How are conflicts between local and global attributes resolved? At least it seems that this distinction could be meaningful.

annevk:
But say you empty global attributes, wouldn't we still want to allow `title` on `abbr`? Or are you saying that's what's unexpected here?

evilpie:
> But say you empty global attributes, wouldn't we still want to allow `title` on `abbr`? Or are you saying that's what's unexpected here?

I am not sure it really make sense to support the usecase of modifying the **built-in safe default configuration** by removing the allowed attributes (in the top-level `attributes`).

mozfreddyb:
Let's consider removing the duplicate entries (but not change algorithm behavior)

PR #278, opened 2025-03-20
Dedupe defaults

otherdaniel:
The safe-default-configuration.txt mirrors the HTML spec, and contains a few attributes both as global attributes as well as per-element attributes (because those elements assign specific semantics to them). Copying both of them into our default config is redundant, and leaves to surprising behaviour if users remove the global attributes (and would still find per-element attributes).

This follows discussion in #271. We remove any dupes from the builtins; but leave the algorithms themselves alone.
This is implemented by more comprehensive dupe-filtering in the script that generates the builtins.

Fixes: #271

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/278.html" title="Last updated on Mar 20, 2025, 12:09 PM UTC (c6d555f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/278/8265187...otherdaniel:c6d555f.html" title="Last updated on Mar 20, 2025, 12:09 PM UTC (c6d555f)">Diff</a>

### DIFF
--- a/builtins/safe-default-configuration.json
+++ b/builtins/safe-default-configuration.json
@@ -248,22 +248,12 @@
     {
       "name": "dfn",
       "namespace": "http://www.w3.org/1999/xhtml",
-      "attributes": [
-        {
-          "name": "title",
-          "namespace": null
-        }
-      ]
+      "attributes": []
     },
     {
       "name": "abbr",
       "namespace": "http://www.w3.org/1999/xhtml",
-      "attributes": [
-        {
-          "name": "title",
-          "namespace": null
-        }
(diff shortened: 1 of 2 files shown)


annevk:
I think both can be surprising and I'm not entirely convinced this is the right call.

I guess this is the more conservative take, so it's probably okay, but it seems a little weird.

mozfreddyb (approved):

mozfreddyb:
Should I merge or do you want to run the build, @otherdaniel ?

PR #280, opened 2025-03-28
Add default() method to retrieve the built-in safe default config.

otherdaniel:
This follows discussion in #233.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -262,6 +262,7 @@ interface Sanitizer {
 
   // Query configuration:
   SanitizerConfig get();
+  static SanitizerConfig default();
 
   // Modify a Sanitizer's lists and fields:
   undefined allowElement(SanitizerElementWithAttributes element);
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -295,6 +296,19 @@ method steps are:
 The <dfn for="Sanitizer" export>get</dfn>() method steps are to return the value of [=this=]'s [=Sanitizer/configuration=].
 </div>
 
+<div algorithm>
+The <dfn for="Sanitizer" export>default</dfn>() method steps are:
```
annevk, line 300:
```suggestion
The static <dfn for="Sanitizer" export>default</dfn>() method steps are:
```

annevk (commented):
I actually think we don't really need this now that we have a mutable `Sanitizer` object and you can get this with `new Sanitizer("default").get()`.

This seems solved.

otherdaniel:
As discussed in #233 and meeting, this method does not seem neccessary. We'll discard this PR.

Issue #154, opened 2022-06-10
(meta) export concepts used in funky element handling 


mozfreddyb:
This is a meta bug for a couple of issues, that we need to fix elsewhere before we can resolve on our end.

- template content is not exported in HTML
- the HTMLHyperLinkElementUtils `protocol` property is also not properly exported. Also, the `protocol` getter steps are hairy and need to be moved into its own algorithm, and then we can refer that.
- For the `form` `action` case we need to invoke the URL parser ourselves. I *think* matching on something like `javascript:` text might not work due to encoding stuff (not entirely sure here).
- The `form action` is the URL of the document, when empty. Are we OK with this? Do we want to think about the case of `javascript` documents here? I'd prefer not to, but mentioning for posterity :)

mozfreddyb:
Most of the above should be resolved by previous work. What's left is:

> The form action is the URL of the document, when empty. Are we OK with this? Do we want to think about the case of javascript documents here? I'd prefer not to, but mentioning for posterity :)

In the meeting, we agreed that we can't help that much once the user is in a `javascript:` document, given that all relative URLs in e.g., `href` are similarly dangerous. We should add a note to that effect.

otherdaniel:
Added a PR #282 for the note.

annevk:
Closing as the note wasn't needed after all.

PR #282, opened 2025-03-28
Add note about relative URLs in javascript:-URL origins.

otherdaniel:
A relative URL in a javascript:-origin may turn into another javascript:-URL. Add a note to alert the reader that Sanitizer API will not handle this case.

This follows discussion in #154.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -522,6 +522,12 @@ The SVG animation case is covered by the
 
 </div>
 
+Note: Within an origin with a `javascript:` URL &mdash; that is, after the
+    browser has already navigated to a `javascript:`-URL &mdash;
+    [=relative-URL strings=] may turn into `javascript:`-URLs, because the
+    origin may be used as the [=base URL=]. These cases are not handled by
+    the Sanitizer.
+
 <div algorithm>
 To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
 1. Let |url| be the result of running the [=basic URL parser=]


mozfreddyb (approved):

index.bs
```diff
@@ -522,6 +522,12 @@ The SVG animation case is covered by the
 
 </div>
 
+Note: Within an origin with a `javascript:` URL &mdash; that is, after the
+    browser has already navigated to a `javascript:`-URL &mdash;
+    [=relative-URL strings=] may turn into `javascript:`-URLs, because the
+    origin may be used as the [=base URL=]. These cases are not handled by
+    the Sanitizer.
```
annevk, line 529: What exactly would the base URL be here? And what is an origin with a javascript: URL? This doesn't seem to be correct.

annevk (changes requested):

index.bs
benbucksch, line 529: If the document origin is a `javacript:` URL, it *MUST NOT* be used as base URL for relative URLs. Such URLs may either be removed entirely from the document, or rewritten into absolute URLs with a safe base URL.

benbucksch (changes requested):

index.bs
benbucksch, line 529:
```suggestion
If the document origin is a `javacript:` URL, it *MUST NOT* be used as base URL for relative URLs. Such URLs may either be removed entirely from the document, or rewritten into absolute URLs with a safe base URL.
```

benbucksch (changes requested):

index.bs
benbucksch: Agreed. I made an different wording proposal below.

annevk: That's also not a correct change.

benbucksch: What's wrong with it?

annevk: I don't think you can end up in such a situation though? Is it really a `javascript:` URL at that point or `about:blank`? You can for the base URL, except HTML already ignores it in such cases.

mozfreddyb: Mh. What do you mean "ignores it"?

otherdaniel: So, after reading specs, it indeed seems that `javascript:`-Urls don't ever become base URLs. I toyed around a bit with this in the browser, and the document created by `javascript:`-navigation uses the previous document's base URL.

Spec-wise, this corresponds to:
* https://html.spec.whatwg.org/multipage/browsing-the-web.html#evaluate-a-javascript:-url, step 11, `URL` parameter,
* ditto, step 16, `about base URL` parameter,
* And https://html.spec.whatwg.org/multipage/urls-and-fetching.html#document-base-url, which references those two values.

So indeed the situation we're warning about here simply doesn't exist in a conforming browser.

I suggest we drop the PR; but I'll wait until the next meeting in case there're any objections.

PR #284, opened 2025-03-31
Don't allow comments and data-attributes when creating a default Sanitizer

### LINKED ISSUE #273: Sanitizer constructor needs to set configuration with safe
Regression from #254. The normal `Sanitizer` constructor does not set the safe parameter.

> 2. Let valid be the return value of [setting](https://wicg.github.io/sanitizer-api/#sanitizer-set-a-configuration) configuration on [this](https://webidl.spec.whatwg.org/#this).

https://wicg.github.io/sanitizer-api/#sanitizer-constructor

> To set a configuration, given a [dictionary](https://webidl.spec.whatwg.org/#dfn-dictionary) configuration, a [boolean](https://infra.spec.whatwg.org/#boolean) safe, and a [Sanitizer](https://wicg.github.io/sanitizer-api/#sanitizer) sanitizer:

https://wicg.github.io/sanitizer-api/#sanitizer-set-a-configuration

It's actually not immediately obvious to me if we should default to `false` or `true`. I think whatever we choose we would end up in a situation where we either break `setHTML(html) == setHTML(html, {sanitizer: new Sanitizer()})` or `setHTMLUnsafe(html) == setHTMLUnsafe(html, {sanitizer: new Sanitizer()})`, because at the construction time we don't know which variant safe/unsafe is going to be called.
We might have discussed this already, though.

evilpie:
1. Changes the parameter of **_set a configuration_** from *safe* to the inverted *allowCommentsAndDataAttributes*.
2. From the constructor don't allow comment/attributes when creating a `"default"` Sanitizer.

Fixes #273

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/284.html" title="Last updated on Apr 2, 2025, 9:58 AM UTC (f94ccc1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/284/3cd0ca8...evilpie:f94ccc1.html" title="Last updated on Apr 2, 2025, 9:58 AM UTC (f94ccc1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -408,7 +408,7 @@ Note: This algorithm works for both {{SetHTMLOptions}} and
 1. If |sanitizerSpec| is a [=dictionary=]:
    1. Let |sanitizer| be a new {{Sanitizer}} instance.
    1. Let |setConfigurationResult| be the result of [=set a configuration=]
-      with |sanitizerSpec| and |safe| on |sanitizer|.
+      with |sanitizerSpec| and [=boolean/not=] |safe| on |sanitizer|.
    1. If |setConfigurationResult| is false, [=throw=] a {{TypeError}}.
    1. Set |sanitizerSpec| to |sanitizer|.
 1. [=Assert=]: |sanitizerSpec| is a {{Sanitizer}} instance.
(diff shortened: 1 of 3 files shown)


annevk (approved):
Looks reasonable to me.

index.bs
```diff
@@ -283,10 +283,13 @@ A {{Sanitizer}} has an associated <dfn for="Sanitizer">configuration</dfn>, a {{
 The <dfn for="Sanitizer" export>constructor</dfn>(|configuration|)
 method steps are:
 
+1. Let |allowCommentsAndDataAttributes| be true.
 1. If |configuration| is a {{SanitizerPresets}} [=string=], then:
     1. [=Assert=]: |configuration| [=is=] {{SanitizerPresets/default}}.
     1. Set |configuration| to the [=built-in safe default configuration=].
-1. Let |valid| be the return value of [=set a configuration|setting=] |configuration| on [=this=].
+    2. Set |allowCommentsAndDataAttributes| to false.
```
otherdaniel, line 290: I wonder if it wouldn't be simpler to just add "comments" and "dataAttributes" keys to the built-in default config in §3.4 Builtins. Then the rest would just do the right thing.

otherdaniel (approved):
Looks good.

One nitpick about whether to change the constant that changes the default, rather than fixing it up after using it. Doesn't make a big difference, though. :)

index.bs
evilpie: We like this approach better, because it puts the whole (default) config into one place. I am going to do this.

evilpie:
I've updated this PR to follow @otherdaniel's suggestion of changing the built-in safe default config (JSON).

evilpie:
Can you merge, please?

Issue #281, opened 2025-03-28
We don't allow allow+remove for global attributes; but we do for per-element attributes.


otherdaniel:
This is split out from #249: 

Config procesing: We currently throw when an element or global attribute is listed in both allow & remove, but not for per-element attributes.

```js
new Sanitizer({elements: ["p"], removeElements: ["p"]})  // throws
new Sanitizer({elements: [{name: "p", attributes: ["id"], removeAttributes: ["id"]}]})  // does not throw.
```

mozfreddyb:
This ought to be closed when #285 is done (like #268)

annevk:
There are some cases that we should think a bit longer about:

* Element allows attribute X that is also a globally removed attribute. Should that throw?
* Global attributes is missing, but element only allows attribute X. Should that throw?

evilpie:
Okay, here is a list of cases with how I (or we) think they should behave combined with what I believe is the current behavior.

### Global attributes and removeAttributes
```
{ attributes: ["foo"], removeAttributes: ["bar"] }
```
Currently: Throws
Suggestion: Keeping throwing

### Element with both attributes and removeAttributes

```
{ elements: [{ name: "p", attributes: ["title"], removeAttributes: ["id"] }], attributes: ["dir", "id"] }
```

Currently: This allows the attributes "title" and "dir" for <p>. "id" was removed even though it is in the global list. 
Suggestion: I think we should try to keep this?

### Element's attributes overlapping with global attributes

```
{ elements: [{name: "p", attributes: ["title"]}], attributes: ["title"] }
```
Currently: Allows "title" on <p>
Suggestion: Throw for overlapping attributes

### Element's removeAttributes overlapping with global removeAttributes

```
{ elements: [{name: "p", removeAttributes: ["id"]}], removeAttributes: ["id"] }
```
Currently: Allows all attributes except "id" ?
Suggestion: Throw for overlapping attributes

### Element's attributes overlapping with global removeAttributes

```
{ elements: [{name: "p", attributes: ["id"]}], removeAttributes: ["id"] }
```
Currently: Allows all attributes _except_ "id" on every element (including `<p>`)
Suggestion: Throw, because "id" can't be allowed for a specific element when it is in the global removeAttributes list?

### Element's removeAttributes overlappign with global attributes

```
{ elements: [{name: "p", removeAttributes: ["id"]}], attributes: ["id", "title"] }
```
[...]

PR #285, opened 2025-04-10
Cleanup sanitize core

### LINKED ISSUE #268: Handling of omitted vs empty lists
For allow lists like `elements` and `attributes` it probably makes sense to handle an empty list differently from an entirely missing list. 

### Example
`setHTML("<b>b</b><i>i</i>", { removeElements: ["i"] })` should probably result in `<i>i</i>`,
while `setHTML("<b>b</b><i>i</i>", { removeElements: ["i"], elements: [] })` results in an empty string.

After #261 is fixed, the Sanitizer could be used to flatten specific elements with `replaceWithChildrenElements`, while either removing all other elements with `elements: []` or keeping all other elements by not using `elements` at all.

### Allow everything to be optional?

The `get` method makes it observable is something is truly optional internally or not. While there AFAIK no observable difference between the config `{ }` and `{ removeElements: [] }`, if we treat `removeElements` as non-optional, the `get()` result would be `{ removeElements: [] }` for both.
If we make `elements` and `attributes` optional, we might consider also making those optional. I don't have a strong opinion either way.

### Alternative

We could also treat `elements: []` as being optional, but I think that would make Sanitizer strictly less powerful.

otherdaniel:
Re-work the handling of empty vs missing lists, along the lines discussed in #268. Empty allow-lists now effectively block everything; while a missing allow-list does nothing. For remove-lists, there is no difference and neither form will remove anything.

Examples:
- `{}` => everything passes
- `{elements: []}` => nothing passes

The error handling is updated to disallow both allow and block lists (for elements & attributes, each). For per-element attributes no such restriction exists; unlike what #281 wants.

Example:
`{ attributes: ["src", "srcset", "alt"], elements: [{name: "img", attributes: ["width", "height"], removeAttributes: ["srcset"]}]}`
This would allow ``, but would drop a `srcset` attribute from the element.

Fixes #268 and #281

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/285.html" title="Last updated on May 13, 2025, 1:11 PM UTC (ce6879e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/285/d4a6cff...otherdaniel:ce6879e.html" title="Last updated on May 13, 2025, 1:11 PM UTC (ce6879e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -456,7 +456,7 @@ template contents). It consistes of these steps:
           |handleJavascriptNavigationUrls|.
       1. Call [=replace all=] with |child|'s [=tree/children=] within |child|.
       1. [=Continue=].
-    1. If |configuration|["{{SanitizerConfig/removeElements}}",&laquo;[]&raquo;]
+    1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=map/with default=] &laquo;[]&raquo;
         [=SanitizerConfig/contains=] |elementName|, or
         |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=] and
         |configuration|["{{SanitizerConfig/elements}}"] does not
@@ -473,18 +473,21 @@ template contents). It consistes of these steps:
     1. [=list/iterate|For each=] |attribute| in |child|'s [=Element/attribute list=]:
       1. Let |attrName| be a {{SanitizerAttributeNamespace}} with |attribute|'s
          [=Attr/local name=] and [=Attr/namespace=].
-      1. If |configuration|["{{SanitizerConfig/removeAttributes}}",&laquo;[]&raquo;] [=SanitizerConfig/contains=] |attrName|,
+      1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/with default=] &laquo;[]&raquo; [=SanitizerConfig/contains=] |attrName|,
          then [=/remove an attribute|remove=] |attribute| from |child|.
-      1. If |configuration|["{{SanitizerConfig/elements}}",&laquo;[]&raquo;]
-          ["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}",&laquo;[]&raquo;]
+      1. Let |elements| be |configuration|["{{SanitizerConfig/elements}}"] [=map/with default=] &laquo;[]&raquo;.
+      1. If |elements|
+          ["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
+          [=map/with default=] &laquo;[]&raquo;
           [=SanitizerConfig/contains=] |attrName|,
           then [=/remove an attribute|remove=] |attribute| from |child|.
       1. If all of the following are false:
-         - |configuration|["{{SanitizerConfig/attributes}}",&laquo;[]&raquo;]
+         - |configuration|["{{SanitizerConfig/attributes}}"]
+            [=map/with default=] &laquo;[]&raquo;
             [=SanitizerConfig/contains=] |attrName|,
-         - |configuration|["{{SanitizerConfig/elements}}",&laquo;[]&raquo;]
-           ["{{SanitizerElementNamespaceWithAttributes/attributes}}",&laquo;[]&raquo;]
-           does not [=SanitizerConfig/contain=] |attrName|, and
+         - elements["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+            [=map/with default=] &laquo;[]&raquo;
+            does not [=SanitizerConfig/contain=] |attrName|, and
          - "data-" is a [=code unit prefix=] of [=Attr/local name=] and
             [=Attr/namespace=] is `null` and
             |configuration|["{{SanitizerConfig/dataAttributes}}"] is true,
@@ -551,12 +554,14 @@ To <dfn for="SanitizerConfig">allow an element</dfn> |element| with a {{Sanitize
     [=map/set=] |configuration|["{{SanitizerConfig/elements}}"] to
     &laquo;[]&raquo;
 1. [=SanitizerConfig/Remove=] |element| from
-    |configuration|["{{SanitizerConfig/elements}}",None].
+    |configuration|["{{SanitizerConfig/elements}}"] [=map/with default=] None.
 1. [=list/Append=] |element| to |configuration|["{{SanitizerConfig/elements}}"].
 1. [=SanitizerConfig/Remove=] |element| from
-    |configuration|["{{SanitizerConfig/removeElements}}",None].
+    |configuration|["{{SanitizerConfig/removeElements}}"] [=map/with default=]
+    None.
 1. [=SanitizerConfig/Remove=] |element| from
-    |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}",None].
+    |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"]
+    [=map/with default=] None.
 
 NOTE: Handling of [=allowElement=] is a little more complicated than the other
     methods, because the element allow list can have per-element allow- and
@@ -583,9 +588,10 @@ To <dfn for="Sanitizer">remove an element</dfn> |element| from a {{SanitizerConf
 1. [=list/Append=] |element| to
     |configuration|["{{SanitizerConfig/removeElements}}"].
 1. [=SanitizerConfig/Remove=] |element| from
-    |configuration|["{{SanitizerConfig/elements}}",None].
+    |configuration|["{{SanitizerConfig/elements}}"] [=map/with default=] None.
 1. [=SanitizerConfig/Remove=] |element| from
-    |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}",None].
+    |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"]
+    [=map/with default=] None.
 
 </div>
 
@@ -602,9 +608,10 @@ from a {{SanitizerConfig}} |configuration|, do:
 1. [=list/Append=] |element| to
     |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"].
 1. [=SanitizerConfig/Remove=] |element| from
-    |configuration|["{{SanitizerConfig/removeElements}}",None].
+    |configuration|["{{SanitizerConfig/removeElements}}"]
+    [=map/with default=] None.
 1. [=SanitizerConfig/Remove=] |element| from
-    |configuration|["{{SanitizerConfig/elements}}",None].
+    |configuration|["{{SanitizerConfig/elements}}"] [=map/with default=] None.
 
 </div>
 
@@ -620,7 +627,8 @@ To <dfn for="Sanitizer">allow an attribute</dfn> |attribute| on a
 1. [=list/Append=] |attribute| to
     |configuration|["{{SanitizerConfig/attributes}}"].
 1. [=SanitizerConfig/Remove=] |attribute| from
-    |configuration|["{{SanitizerConfig/removeAttributes}}",None]
+    |configuration|["{{SanitizerConfig/removeAttributes}}"]
+    [=map/with default=] None.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -468,24 +472,28 @@ template contents). It consistes of these steps:
...
-         - |configuration|["{{SanitizerConfig/attributes}}"] [=list/exists=] and
-           [=SanitizerConfig/contains=] |attrName|
-         - |configuration|["{{SanitizerConfig/elements}}"]["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
-           [=SanitizerConfig/contains=] |attrName|
+      1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=]
+          and [=SanitizerConfig/contains=] |attrName|,
+         then [=/remove an attribute|remove=] |attribute| from |child|.
+      1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=] and
+          |configuration|["{{SanitizerConfig/elements}}"]["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
+          [=map/exists=] and [=SanitizerConfig/contains=] |attrName|,
```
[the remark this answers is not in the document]
otherdaniel: Done, by seperating the clauses with `, and if`.

evilpie:
Indirectly related to this, but I don't want to forget about this. 

Don't we also need to handle missing properties in [set a configuration](https://wicg.github.io/sanitizer-api/#sanitizer-set-a-configuration) when handling the _input_ config? We just access `configuration["elements" etc.

Similarly in [allow an element](https://wicg.github.io/sanitizer-api/#sanitizerconfig-allow-an-element) etc. we are just appending/removing from potentially missing lists.

evilpie:
Oh, and the way it's currently specified, we can't actually tell the difference between `{ elements: [] }` and `{}`, because we don't differentiate the two cases properly in _set a configuration_.

otherdaniel:
Thanks for the review!

> Similarly in [allow an element](https://wicg.github.io/sanitizer-api/#sanitizerconfig-allow-an-element) etc. we are just appending/removing from potentially missing lists.

Fixed this in the latest push.

(Other fixes are still coming...)

index.bs
```diff
@@ -640,16 +669,42 @@ To <dfn for="SanitizerConfig">remove unsafe</dfn> from a |configuration|, do thi
...
-1. [=list/iterate|For each=] |attribute| of |configuration|["{{SanitizerConfig/removeAttributes}}"] do:
-  1. Call [=Sanitizer/remove an attribute=] with |attribute| and |sanitizer|'s [=Sanitizer/configuration=].
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+  1. Set |sanitizer|'s |configuration|["{{SanitizerConfig/elements}}"]
+      to &laquo;[]&raquo;
+  1. [=list/iterate|For each=] |element| of
+      |configuration|["{{SanitizerConfig/elements}}"]:
+    1. Call [=allow an element=] with |element| and
+        |sanitizer|'s [=Sanitizer/configuration=].
+1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=map/exists=]:
```
evilpie, line 679: We shouldn't need the exists checks anymore, now that they are always initialized to empty lists.

evilpie:
> "set a configuration" now explicitly initializes any list that exists in the input. So absent and empty lists should now be accurately reflected in the internal config.

I would have actually preferred to always initialize everything but elements/attributes, but I can also totally see how only having those be optional would be confusing. (It does emphasize the difference between missing and empty being special for these though ...)

I think we are still missing an initialization of "attributes" and "remove attributes" in **canonicalize a sanitizer element with attributes**.

index.bs
```diff
@@ -468,24 +472,27 @@ template contents). It consistes of these steps:
...
+          and [=SanitizerConfig/contains=] |attrName|,
+         then [=/remove an attribute|remove=] |attribute| from |child|.
+      1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=], and
+          if |configuration|["{{SanitizerConfig/elements}}"]["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
+          [=map/exists=] and [=SanitizerConfig/contains=] |attrName|,
+          then [=/remove an attribute|remove=] |attribute| from |child|.
+      1. If all of the following are false,
+          then [=/remove an attribute|remove=] |attribute| from |child|.
+         - |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=] and
+           does not [=SanitizerConfig/contain=] |attrName|
```
evilpie, line 485: Isn't "does not" wrong? If the attribute is contained in SanitizerConfig/attributes (i.e. it should be allowed) this expression would result in `false`, which would mean removal.

This step might be the most complicated in the whole specification, we should try to brainstorm something to simplify it.

otherdaniel: > Isn't "does not" wrong? If the attribute is contained in SanitizerConfig/attributes (i.e. it should be allowed) this expression would result in `false`, which would mean removal.

Yes, I think so. I'll update the PR.

> This step might be the most complicated in the whole specification, we should try to brainstorm something to simplify it.

I also agree. I tried to keep it simple, but... this is the result. :-)

Alternatives or improvements I can think of:
- Be more aggressive about factoring out common sub-expression. In particular, `if map/exists and SanitizerConfig/contains` occurs a lot.
- This is basically an if-then with 3 sub-conditions. One could either move all the conditions into a seperate algorithm. (Which I'm usually not that fond of, because that mostly only introduces an indirection, rather than a proper abstraction.)
- We could resolve the 3-condition  if-then into three seperate if-thens. That would repeat the action 3 times, though.
- We could resolve the 3-condition if-then into three seperate if-thens, which skip over the remaining steps of this sequence of steps. I.e., "If [...], then skip over the remaining steps in this sequence," sort of like `continue` or `break`does for loops. The infra spec defines suitable languages for loops, but not for overly long if-thens. I think we can just describe it in English, though.

wdyt?

index.bs
```diff
@@ -450,12 +450,16 @@ template contents). It consistes of these steps:
...
-    1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=SanitizerConfig/contains=] |elementName|:
+    1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"]
+        [=map/exists=] and [=SanitizerConfig/contains=] |elementName|:
       1. Call [=sanitize core=] on |child| with |configuration| and
           |handleJavascriptNavigationUrls|.
       1. Call [=replace all=] with |child|'s [=tree/children=] within |child|.
       1. [=Continue=].
-    1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=SanitizerConfig/contains=] |elementName|, or if |configuration|["{{SanitizerConfig/elements}}"] is not [=list/empty=] and does not [=SanitizerConfig/contain=] |elementName|:
+    1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=map/exists=] and
+        [=SanitizerConfig/contains=] |elementName|, or if
```
annevk, line 460: I see you sometimes use a nested "if". The usual specification style does not have that.

index.bs
annevk: I would make at least one change. Write it like

> If all of the following are false:
> * a is true;
> * b is true; and
> * c is true,
>
> then ...

to make it flow a bit better (and be more aligned with Infra). (Might be harder with Markdown, not sure.)

index.bs
```diff
@@ -537,11 +544,16 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
 
-1. Set |element| to the result of [=canonicalize a sanitizer element with attributes=] with |element|.
-1. [=SanitizerConfig/Remove=] |element| from |configuration|["{{SanitizerConfig/elements}}"].
-1. [=list/Append=] |element| to |configuration|["{{SanitizerConfig/elements}}"].
-1. [=SanitizerConfig/Remove=] |element| from |configuration|["{{SanitizerConfig/removeElements}}"].
-1. [=SanitizerConfig/Remove=] |element| from |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"].
+1. Set |element| to the result of [=canonicalize a sanitizer element with
+    attributes=] with |element|.
+1. [=SanitizerConfig/Remove=] |element| from the "{{SanitizerConfig/elements}}"
+    key of |configuration|.
```
annevk, line 550: This notation is very confusing. Should this be

> Remove _element_ from _configuration_["elements"]

?

index.bs
```diff
@@ -755,21 +807,40 @@ if there exists an |entry| of |list| that is an [=ordered map=], and where
...
+1. Let |list| be |configuration|[|key|].
 1. If |list| [=SanitizerConfig/contains=] |name|, then return.
 1. [=list/Append=] |name| to |list|.
 
 </div>
 
+<div algorithm>
+To determine the <dfn for="SanitizerConfig">size</dfn> of a [=/list=] |list|:
+
+1. If |list| does not [=map/exist=], return None.
```
annevk, line 839: Using map/exists on a list doesn't make sense.

index.bs
otherdaniel: It is confusing indeed.

The problem is that all those keys are optional and might not be there. The intent is that I can move the existence check into the algorithm that get's called: If *configuration* doesn't actually have an *"elements"*, then *Remove element from configuration["elements"]* would run into the assertion at the beginning of *[=map/get the value of an entry=]*. Here, I handle this by moving the map operation into the algorithm, and can do the existence-check at the beginning of *remove*.

One alternative would be to just do the existence check plus default value at every calling site. That strikes me as very repetetive.

Ideal would be if there'd be an *a[b]* operator that would return *None* (or somesuch). Then the calling algorithm could just treat *None* in whichever fashion it likes.

otherdaniel: I've prepared a version based on the latter, based on an *a[b,default]* operation. Not sure how much that improves things, and whether that's at all palatable for eventual inclusion into HTML. On the pro side, it helps with this problem; with the *size* problem below, and occurs again in the core algorithm issue above. On the con side, I'm inventing another new notation. :-/

Please let my know what you all think. I'm still unsure how to proceed.

index.bs
otherdaniel: Yes. The intent was that this gets called with *configuration["bla"]* and I'm hoisting the existance check into the caller. I don't think that was valid. I've now prepared a version that's explicit.

index.bs
otherdaniel: I followed the 'flow' suggestion here, and simplified the expressions a little bit by merging existance check + lookup into one operation.

index.bs
otherdaniel: Removed "or if" and "and if".

index.bs
evilpie: I don't think I like that `size` can return None. We should probably do something more explicit for "set  a configuration"?

Like, roughly:

*Did list change?* 
1. If a and b is some, return number of items in a and b are identical.
2. Return true if both a and b are none
3. Return false

evilpie:
Could we move the DocumentType change to another PR?

otherdaniel:
> Could we move the DocumentType change to another PR?

Done: #289 Will also remove those bits from this PR.

evilpie:
(Sorry, All these comments are getting confusing)

Coming back to dreaded attributes removal condition. I still don't think it really works. Imagine 1. `{ attributes: [] }` vs. 2. `{  }`. 1. should probably remove all attributes, while 2. keeps all of them?

index.bs
```diff
@@ -782,6 +871,24 @@ To determine <dfn for="boolean">not</dfn> of a [=boolean=] |bool|,
...
+
+</div>
+
+We can also denote [=maybe get the value|maybe getting the value=] using an indexing syntax with two parameters, by providing a [=map/key=] and a default value seperated by comma, inside square brackets directly following a map.
+
+<div class=example>
+`map["test",None]` would return
+`"ok"` for a [=map=] &laquo;["test"&rightarrow;"ok"]&raquo;,
+and `None` for a [=map=] &laquo;["ok"&rightarrow;"test"]&raquo;.
+</div>
```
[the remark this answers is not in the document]
otherdaniel: Done: https://github.com/whatwg/infra/issues/668

evilpie:
I think we can close this in favor of #296 already.

Issue #292, opened 2025-05-20
The modifier methods produce SanitizerConfigs that aren't valid


evilpie:
Consider this (pretend some of the fixes in #285 are included):

```
var sanitizer = new Sanitizer({ removeAttributes: ["id"] });
sanitizer.allowAttribute(["id"]) 
sanitizer.get() // { attributes: ["id"], removeAttributes: [] }
new Sanitizer({ attributes: ["id"], removeAttributes: [] }) // throws (attributes and removeAttributes)
```

These things get 100% more complicated when we actually try to implement some of my suggestions from #281. Like should we be checking allow attributes-per-element again after modifying the global attributes etc.? What happens when allowing an attribute that isn't in the `removeAttributes` list (without an `attributes` list of course)?

It seems like the modifiers make it very easy to create a Sanitizer that is actually quite hard to reason about, especially compared to the massive restrictions proposed for the constructor in  #281. In fact I would say our current definition of these methods don't really work together with how I envision the constructor should work.

annevk:
Yeah, it seems we'll have to approach this differently, but the number of conditions to account for might make this quite involved.
```js
const sanitizer = new Sanitizer({ removeAttributes: ["id"] });
sanitizer.allowAttribute(["id"]); 
```
Perhaps what you really want to end up with here is `{ removeAttributes: [] }`. But when what about:
```js
const sanitizer = new Sanitizer({ removeAttributes: ["title"] });
sanitizer.allowAttribute(["id"]); 
```
Should that no-op or throw? Perhaps it should return a boolean as to whether the operation succeeded? Typically you don't really want these kind of things to throw.

But working that out for each manipulation method and possible configuration is going to be some work.

mozfreddyb:
I hope I'm not kicking a hornet's nest here, but how about we consider moving these methods to a v2 (or v1.1). I don't want us to spin off another endless discussion, while we still haven't gathered real user feedback in pre-release browsers.

What do others think?

otherdaniel:
> I hope I'm not kicking a hornet's nest here, but how about we consider moving these methods to a v2 (or v1.1). I don't want us to spin off another endless discussion, while we still haven't gathered real user feedback in pre-release browsers.

I share the worry that we might be going in circles here; but at least so far the discussion seems to be going forward. :)

[...]

PR #293, opened 2025-05-21
Draft: Heavily limit the possible Sanitizer configurations

evilpie:
This is still a total draft, but this mostly shows my suggested behavior of #281. This is pretty much a replacement for #285, except of course it's not yet actually something we have discussed in any form.
I haven't yet removed the modifier methods (#292), but this doesn't really work without doing that or spending more time coming up with new definitions for those methods.

I think I also like having the exists checks be explicit.

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/293.html" title="Last updated on May 21, 2025, 11:04 AM UTC (94676e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/293/50ed6b6...evilpie:94676e9.html" title="Last updated on May 21, 2025, 11:04 AM UTC (94676e9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -453,14 +453,19 @@ template contents). It consistes of these steps:
   1. Otherwise:
     1. Let |elementName| be a {{SanitizerElementNamespace}} with |child|'s
        [=Element/local name=] and [=Element/namespace=].
-    1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=SanitizerConfig/contains=] |elementName|:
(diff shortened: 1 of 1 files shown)


evilpie:
Ups, I forgot about `removeUnsafe` while working on this 😆. Instead of trying to modify the config on the fly, it would probably be easier to just remove the handful of elements and attributes explicitly in the core algorithm.

evilpie:
Let's close this. I was primarily trying figure out how to write the config verification just in the Sanitizer constructor. I am pretty much convinced that we should instead continue to do the canonicalization and implicitly verification using the modifier methods instead.

Issue #294, opened 2025-05-28
Element attributes: Enforcing a strict non-ambiguous config vs. using a well defined order of operations


evilpie:
In this issue I am trying to capture to largest remaining point of contention that we have identified in issues #281 and #292.

Specially there is disagreement on how more complicated uses of element attributes should work. These usually involve having both a global attributes/removeAttributes and per-element attributes lists that all overlap in some way. One school of thought would be to reject using these maybe more complicated configs, while the other school of thought wants to avoid the associated overhead and complexity with doing this verification by instead defining clear semantics for how these conflicts would be resolved.

I don’t think we have very clearly defined what the order of operations would be, but I imagine it would be either this:

1. Check if the there are attributes lists for this element tag
   1. If the attribute name is contained in removeAttributes remove
   2. If the attribute name is contained in attributes allow (and return?)
2. Check if (global) removeAttributes exists
   1. If it contains the attribute, remove
3. Else
   1. If (global) attributes doesn’t contain attributes, remove

Alternatively I would imagine we could also always handle the global removeAttributes before the per-element attributes (So Step 2 before Step 1.). This would make it impossible to allow an attribute per element that was globally removed. 

Let’s look at some examples:

```
{ elements: [{ name: "div", attributes: ["id"], removeAttributes: ["id"] }], attributes: [] }
```

Config: throws, duplication
Order of operations: Remove before allow

```
{ elements: [{ name: "div", attributes: ["id"] }], attributes: [] }
```

Config: no conflict
Order of operations: allow id only for div

```
{ elements: [{ name: "div", attributes: ["id"] }], removeAttributes: ["id"] }
```

Config: Allow id for div or throw? (This is the case that I find personally confusing)
Order of operations: Depends on what we specify, because either per-element attribute list always trumps or the global removeAttributes always wins.

```
{ elements: [{ name: "div", removeAttributes: ["id"] }], attributes: ["id"] }
```
Config: Don’t allow id for div, allow id for all other elements
Order of operations: Remove per-element always wins

annevk:
I don't think throwing for per-element allow attributes combined with global remove attributes works, because we don't require each element to be the same. Some elements might have local allow attributes, others might not.

I think for the second scenario (local allow and local remove for the same element, and global allow) 1 is more elegant as well, so going with that certainly works for me.

evilpie:
[...]

PR #296, opened 2025-06-03
Re-work configuration handling and sanitize-core.

otherdaniel:
This PR is meant to capture our recent discussion. It surely needs more editing; but I think the general structure is there. I'd be particularly interested in quick feedback on whether this accurately captures the discussion consensus.

-------------------
This re-writes config handling and sanitize-core based on 2025-05-28 and 2025-05-14 meeting discussion. This would replace #285.

Since we couldn't quite agree on whether redundent global attribute and per-element attribute definitions should be considered an error, I have tried to cover both. Where they differ, a statement "IF DISALLOW REDUNDANT PER-ELEMENT-ATTRIBUTES" contains the additional checks needed to disallow this.

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/296.html" title="Last updated on Sep 12, 2025, 3:21 PM UTC (f1e0dae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/296/50ed6b6...otherdaniel:f1e0dae.html" title="Last updated on Sep 12, 2025, 3:21 PM UTC (f1e0dae)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -411,10 +411,10 @@ Several conditions need to hold for a configuration to be valid:
     {{SanitizerConfig/attributes}} allow-list. The {{SanitizerConfig/dataAttributes}}
     attribute is only allowed when a {{SanitizerConfig/attributes}} list is used.
 - Duplicate entries between different global lists:
-  - There are no duplicate entries (i.e. no same elements) between
+  - There are no duplicate entries (i.e., no same elements) between
     {{SanitizerConfig/elements}}, {{SanitizerConfig/removeElements}}, or
     {{SanitizerConfig/replaceWithChildrenElements}}.
-  - There are no duplicate entries (i.e. no same attributes) between
+  - There are no duplicate entries (i.e., no same attributes) between
     {{SanitizerConfig/attributes}} or {{SanitizerConfig/removeAttributes}}.
 - Duplicate entries on the same element:
   - There are no duplicate entries between {{SanitizerElementNamespaceWithAttributes/attributes}}
@@ -572,7 +572,7 @@ Note: [=set a configuration|Setting a configuration=] from a [=dictionary=] will
 normalization. In particular, if both allow- and remove-lists are missing, it will interpret this
 as an empty remove-list. So `{}` itself is not a [=SanitizerConfig/valid=] configuration, but it
 will be normalized to `{removeElements:[],removeAttributes:[]}`, which is. This normalization step
-was chosen in order to have a missing dictionary be consistent with an empty one, i.e. to have
+was chosen in order to have a missing dictionary be consistent with an empty one, i.e., to have
 `setHTMLUnsafe(txt)` be consistent with `setHTMLUnsafe(txt, {sanitizer: {}})`.
 
 # Algorithms # {#algorithms}


index.bs
```diff
@@ -535,66 +617,147 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+    1. [=Sanitizer/Add an item=] |attribute| to |configuration|
+        list "{{SanitizerConfig/attributes}}".
+    1. **IF DISALLOW REDUNDANT PER-ELEMENT-ATTRIBUTES:**
+        1. If |configuration|[{{SanitizerConfig/elements}}] [=map/exists=]:
+            1. [=list/iterate|For each=] |element| in |configuration|[{{SanitizerConfig/elements}}]:
+                1. If |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] [=map/exists=]:
+                    1. [=Sanitizer/Remove an item=] |attribute| from
+                        |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}]
+    1. If the [=Sanitizer/add an item=] call returned false or
```
evilpie: I don't think we want to do this implicitly, it seems like it could be quite easy to forget to keep track of one of the return values.

index.bs
```diff
@@ -535,66 +617,147 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+            1. [=list/iterate|For each=] |attr| in
+                |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}]:
+                1. [=Remove an item=] |attr| from
+                    |configuration|[{{SanitizerElementNamespaceWithAttributes/attributes}}].
+        1. If |element| [=map/contains=] {{SanitizerElementNamespaceWithAttributes/removeAttributes}}
+            and |configuration| [=map/contains=] {{SanitizerConfig/removeAttributes}}:
+            1. [=list/iterate|For each=] |attr| in
+                |element|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}]:
+                1. [=Remove an item=] |attr| from
+                    |configuration|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}].
```
evilpie: For element `removeAttributes` we also want to fail when it's not in the global `elements` (if it exists)

otherdaniel: Added.

index.bs
otherdaniel: I think I agree; but haven't done the edits yet.

index.bs
```diff
@@ -643,60 +821,75 @@ To <dfn for="SanitizerConfig">remove unsafe</dfn> from a |configuration|, do thi
...
-  1. Call [=replace an element with its children=] with |element| and |sanitizer|'s [=Sanitizer/configuration=].
-1. [=list/iterate|For each=] |attribute| of |configuration|["{{SanitizerConfig/attributes}}"] do:
-  1. Call [=allow an attribute=] with |attribute| and |sanitizer|'s [=Sanitizer/configuration=].
-1. [=list/iterate|For each=] |attribute| of |configuration|["{{SanitizerConfig/removeAttributes}}"] do:
-  1. Call [=Sanitizer/remove an attribute=] with |attribute| and |sanitizer|'s [=Sanitizer/configuration=].
+1. Let |result| be a new {{SanitizerConfig}}.
+1. [=Comment=]: Because non-existing keys are interpreted differently from keys with
+    empty lists, we need to carefully setup the lists in |result| to match
+    what we find in |configuration| and to match our validity criteria.
+1. If any of the following conditions is true, return false:
```
evilpie: I think we should just make this two steps.

index.bs
```diff
@@ -643,60 +821,75 @@ To <dfn for="SanitizerConfig">remove unsafe</dfn> from a |configuration|, do thi
...
+    - |configuration| [=map/contains=] {{SanitizerConfig/elements}} and
+      |configuration| [=map/contains=] {{SanitizerConfig/removeElements}}
+    - |configuration| [=map/contains=] {{SanitizerConfig/attributes}} and
+      |configuration| [=map/contains=] {{SanitizerConfig/removeAttributes}}
+1. If |result| does not [=map/contain=] {{SanitizerConfig/elements}}
```
evilpie: This should not contain `elements` *and* `removeElements`.

evilpie:
I think in *set a configuration* we should first handle `attributes` and `removeAttributes` and afterwards `elements`, it avoids having to go through all per-element attributes for every global attribute. While I think this shouldn't really be observable from a spec perspective, I think most implementation will want to produce error messages that might make this difference observable in practice.

index.bs
```diff
@@ -535,66 +621,154 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+1. Set |element| to the result of [=canonicalize a sanitizer element with
+    attributes=] with |element|.
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+    1. [=Add an item=] |element| to |configuration| list "{{SanitizerConfig/elements}}".
+    1. [=Remove an item=] |element| from |configuration| list
+        "{{SanitizerConfig/replaceWithChildrenElements}}"
+    1. **IF DISALLOW REDUNDANT PER-ELEMENT-ATTRIBUTES**:
+        1. If |element| [=map/contains=] {{SanitizerElementNamespaceWithAttributes/attributes}}
+            and |configuration| [=map/contains=] {{SanitizerConfig/attributes}}:
+            1. [=list/iterate|For each=] |attr| in
```
evilpie: As mentioned in the meeting: We need to iterate on the global lists and change the local lists.

index.bs
```diff
@@ -368,6 +368,66 @@ dictionary SanitizerConfig {
...
+        {{SanitizerElementNamespaceWithAttributes/removeAttributes}} if it's also listed in
+        the global {{SanitizerConfig/attributes}} list.
+
+
+<div algorithm>
+A {{SanitizerConfig}} is <dfn for=SanitizerConfig>valid</dfn> if all of the following conditions
+hold:
+
+1. The config has an {{SanitizerConfig/elements}} or an {{SanitizerConfig/removeElements}}
+    [=map/key=], or neither, but not both.
```
evilpie:
```suggestion
    [=map/key=], but not both.
```

index.bs
```diff
@@ -471,20 +541,36 @@ template contents). It consistes of these steps:
...
-           [=SanitizerConfig/contains=] |attrName|
-         - "data-" is a [=code unit prefix=] of [=Attr/local name=] and
-            [=Attr/namespace=] is `null` and
-            |configuration|["{{SanitizerConfig/dataAttributes}}"] is true
+      1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=] and
+         |configuration|["{{SanitizerConfig/removeAttributes}}"] [=SanitizerConfig/contains=]
+         |attrName|:
+         1. [=/Remove=] |attribute| from |child|.
+         1. [=Continue=].
+      1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=] and
```
evilpie: We need to do this after checking the per-element `attributes` list, because it might contain it, while the global `attributes` list doesn't.

index.bs
```diff
@@ -453,12 +514,21 @@ template contents). It consistes of these steps:
...
-    1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=SanitizerConfig/contains=] |elementName|:
+    1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/exists=] and
+        |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"]
+        [=SanitizerConfig/contains=] |elementName|:
       1. Call [=sanitize core=] on |child| with |configuration| and
           |handleJavascriptNavigationUrls|.
       1. Call [=replace all=] with |child|'s [=tree/children=] within |child|.
       1. [=Continue=].
-    1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=SanitizerConfig/contains=] |elementName|, or if |configuration|["{{SanitizerConfig/elements}}"] is not [=list/empty=] and does not [=SanitizerConfig/contain=] |elementName|:
+    1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=map/exists=] and
```
evilpie, line 669: It's not strictly necessary for correctness, but I think we should write it in a way that makes it obvious that we always have either `elements` or `removeElements`.

1. If configruation["elements"] exists:
1.1. Remove child, if not elementName in configuration["elements"]
2. Otherwise,
2.1. Remove child, if elementName in configuration["removeElements"]

otherdaniel:
I have updated the "Configuration Invariants" chapter (but not the actual algorithms) to match our discussion from last Wednesday: https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/296.html#invariants 

I'd be happy for a rough review whether this matches what we discussed. If so, I'll update the algorithm parts next.

evilpie:
Adding these tables to the spec is a good idea.

> An attribute is allowed, if it’s in the local allow list, and not in the global remove list. No duplicate entries between global remove and local allow lists are allowed. 

I would word this as 

> An attribute is only allowed if it’s in the local allow list. No duplicate entries between global remove and local allow lists are allowed. 

to make it clearer that this is a strict list. I don't think we need "not in the global remove list" in the first sentences, as this is already enforced by the second.

> **local attributes**: [Custom data attributes](https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute) are allowed. No [custom data attributes](https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute) may be listed in any allow-list, as that would mean a duplicate entry.

I misunderstood this the first time when reading. Maybe "All custom data attributes are allowed..." would help? Not sure.

index.bs
otherdaniel: > Adding these tables to the spec is a good idea.
> 
> > An attribute is allowed, if it’s in the local allow list, and not in the global remove list. No duplicate entries between global remove and local allow lists are allowed.
> 
> I would word this as
> 
> > An attribute is only allowed if it’s in the local allow list. No duplicate entries between global remove and local allow lists are allowed.
> 
> to make it clearer that this is a strict list. I don't think we need "not in the global remove list" in the first sentences, as this is already enforced by the second.

Done. I also added an additional sentence that the global remove list might apply elsewhere.

> > **local attributes**: [Custom data attributes](https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute) are allowed. No [custom data attributes](https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute) may be listed in any allow-list, as that would mean a duplicate entry.
> 
> I misunderstood this the first time when reading. Maybe "All custom data attributes are allowed..." would help? Not sure.

Done.

otherdaniel:
> Adding these tables to the spec is a good idea.
> 
> > An attribute is allowed, if it’s in the local allow list, and not in the global remove list. No duplicate entries between global remove and local allow lists are allowed.
> 
> I would word this as
> 
> > An attribute is only allowed if it’s in the local allow list. No duplicate entries between global remove and local allow lists are allowed.
> 
> to make it clearer that this is a strict list. I don't think we need "not in the global remove list" in the first sentences, as this is already enforced by the second.

Done. Also added a sentence that the global remove list might apply elsewhere.

> > **local attributes**: [Custom data attributes](https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute) are allowed. No [custom data attributes](https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute) may be listed in any allow-list, as that would mean a duplicate entry.
> 
> I misunderstood this the first time when reading. Maybe "All custom data attributes are allowed..." would help? Not sure.

Done.

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+    1. The [=set/difference=] of |config|[{{SanitizerConfig/elements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
+    1. The [=set/difference=] of |config|[{{SanitizerConfig/removeElements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
+    1. The [=set/difference=] of |config|[{{SanitizerConfig/attributes}}] and
+        |config|[{{SanitizerConfig/removeAttributes}}] is [=list/empty=].
+    1. If |config|[{{SanitizerConfig/attributes}}] [=map/exists=]:
+        1. [=list/iterate|For any=] |element| in |config|[{{SanitizerConfig/elements}}]:
+            1. The [=set/difference=] of |config|[{{SanitizerConfig/attributes}}] and
+                |config|[|element|][{{SanitizerElementNamespaceWithAttributes/attributes}}] is
```
evilpie: All of these need to be `|config|[{{SanitizerConfig/elements}}][|element|] ...` or maybe `|element|[...]` ?

otherdaniel: Done. (Used the |element| form.)

index.bs
```diff
@@ -264,16 +266,16 @@ interface Sanitizer {
...
-  undefined setDataAttributes(boolean allow);
+  boolean allowElement(SanitizerElementWithAttributes element);
+  boolean removeElement(SanitizerElement element);
+  boolean replaceElementWithChildren(SanitizerElement element);
+  boolean allowAttribute(SanitizerAttribute attribute);
+  boolean removeAttribute(SanitizerAttribute attribute);
+  boolean setComments(boolean allow);
+  boolean setDataAttributes(boolean allow);
 
   // Remove markup that executes script. May modify multiple lists:
```
annevk: "May modify multiple lists" applies to the methods above as well. Perhaps it would be clearer to not state anything about that.

otherdaniel:
I think this latest version represents the current group thinking; and I hope I've incorporated all feedback.

This ends up being more complicated than what is proposed in [#294](https://github.com/WICG/sanitizer-api/issues/294#issuecomment-2965792542), but largely because there's a few more edge cases this needs to handle. (E.g. neither `attributes` or `removeAttributes` present; or the global `removeAttributes` + local `attributes` case.)

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
 };
 </pre>
 
+## Configuration Invariants ## {#invariants}
+
+An empty configuration allows everything (when called with the "unsafe" methods).
```
evilpie: > (when called with the "unsafe" methods)

I think this is slightly unclear, especially because `removeUnsafe` exists. Maybe:

> (when used as a configuration for "unsafe" methods like `setHTMLUnsafe`)

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+  - There are no duplicate entries (i.e. no same elements) between
+    {{SanitizerConfig/elements}}, {{SanitizerConfig/removeElements}}, or
+    {{SanitizerConfig/replaceWithChildrenElements}}.
+  - There are no duplicate entries (i.e. no same attributes) between
+    {{SanitizerConfig/attributes}} or {{SanitizerConfig/removeAttributes}}.
+
+
+The {{SanitizerConfig/elements}} element allow-list can also specify allowing or removing
+attributes for a given element. This is meant to mirror [[HTML]]'s structure, which knows
+both [=/global attributes=] as well as local <a spec=dom>attributes</a> that apply to a
```
evilpie, line 425: The link to [attributes](https://dom.spec.whatwg.org/#concept-attribute) is confusing to me. Global attributes are still attributes when applied to a specific element.

I think we should not use the terms global or local attribute at all. Global attributes as defined by HTML are different from what you call "global" attributes, i.e. attributes in top-level `attributes` list. (Elements in that list don't necessarily need to be global attributes as defined by HTML)

I don't have a good suggestion for "global" attributes, but I have been calling local attributes "per-element" attributes.

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+specific element. For {{SanitizerConfig|Sanitizer configurations}}, it is assumed that
+configuration authors would prefer either all-global or all-local attribute declarations.
```
evilpie: Why is that assumed? Even the default config has both "local" and "global" attributes.

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+<table class=data>
+<tr>
+ <td></td>
+ <th>global {{SanitizerConfig/attributes}}</th>
+ <th>global {{SanitizerConfig/removeAttributes}}</th>
+</tr>
+<tr>
+ <th>local {{SanitizerElementNamespaceWithAttributes/attributes}}</th>
+ <td class=long>An attribute is allowed if it matches either list. No duplicates are allowed.</td>
+ <td class=long>An attribute is allowed if it's in the local allow list.
```
evilpie: Maybe _only_ allowed to emphasize?

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+    1. If |config|[{{SanitizerConfig/attributes}}] [=map/exists=]:
+        1. [=list/iterate|For any=] |element| in |config|[{{SanitizerConfig/elements}}]:
+            1. The [=set/difference=] of |config|[{{SanitizerConfig/attributes}}] and
+                |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] is
+                [=list/empty=].
+            1. The [=set/difference=] of |config|[{{SanitizerConfig/attributes}}] and
+                |element|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}] is
+                [=list/empty=].
+            1. If {{SanitizerConfig/dataAttributes}} is true:
+                1. Neither |config|[{{SanitizerConfig/attributes}}] nor
```
evilpie: We need to check `|config|[{{SanitizerConfig/attributes}}]` outside the `For any` loop, in case `elements` is empty.

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+
+Putting these rules in words:
+
+* Duplicates and interactions between global and local lists:
+  * If a global {{SanitizerConfig/attributes}} allow list exists, then all element's local lists:
+    * If a local {{SanitizerElementNamespaceWithAttributes/attributes}} allow lists exists,
+        there may be no duplicate entries between these lists.
+    * If a local {{SanitizerElementNamespaceWithAttributes/removeAttributes}} remove lists exists,
+        then all its entries must also be listed in the global {{SanitizerConfig/attributes}}
+        allow list.
```
evilpie, line 507: Could/should we mention somewhere here or somewhere else that this implies that there can be no duplicates between "local" `attributes` and `removeAttributes` lists?

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+A {{SanitizerConfig}} |config| is <dfn for=SanitizerConfig>valid</dfn> if all of the following
+conditions hold:
+
+1. The |config| has an {{SanitizerConfig/elements}} or an {{SanitizerConfig/removeElements}}
+    [=map/key=], or neither, but not both.
+1. The |config| has an {{SanitizerConfig/attributes}} or an {{SanitizerConfig/removeAttributes}}
+    [=map/key=], or neither, but not both.
+1. Assuming that all element names have run through [=canonicalize a sanitizer element=] algorithm,
+    and all attribute names have run through [=canonicalize a sanitizer attribute=] algorithm.
+    1. The [=set/difference=] of |config|[{{SanitizerConfig/elements}}] and
```
evilpie: Remove this.

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+        |config|[{{SanitizerConfig/removeElements}}] is [=list/empty=].
+    1. The [=set/difference=] of |config|[{{SanitizerConfig/elements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
+    1. The [=set/difference=] of |config|[{{SanitizerConfig/removeElements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
+    1. The [=set/difference=] of |config|[{{SanitizerConfig/attributes}}] and
```
evilpie: Remove this.

index.bs
```diff
@@ -368,6 +370,160 @@ dictionary SanitizerConfig {
...
+        |config|[{{SanitizerConfig/removeAttributes}}] is [=list/empty=].
+    1. If |config|[{{SanitizerConfig/attributes}}] [=map/exists=]:
+        1. [=list/iterate|For any=] |element| in |config|[{{SanitizerConfig/elements}}]:
+            1. The [=set/difference=] of |config|[{{SanitizerConfig/attributes}}] and
+                |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] is
+                [=list/empty=].
+            1. The [=set/difference=] of |config|[{{SanitizerConfig/attributes}}] and
```
evilpie: Isn't this wrong?

index.bs
```diff
@@ -541,74 +730,193 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+                1. [=list/Remove=] |attr| from |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
+        1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=]:
+            1. [=map/iterate|For each=] |attr| in |element| [=map/contains=]
+                "{{SanitizerElementNamespaceWithAttributes/attributes}}":
+                1. If |configuration|["{{SanitizerConfig/attributes}}"] does not
+                    [=map/contain=] |attr|:
+                    1. [=list/Remove=] |attr| from |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
+    1. If |element| [=map/contains=] "{{SanitizerElementNamespaceWithAttributes/removeAttributes}}":
+        1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+            1. [=map/iterate|For each=] |attr| in |element| [=map/contains=]
```
evilpie: Spurious `[=map/contains=]` ?

index.bs
```diff
@@ -541,74 +730,193 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+    |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"].
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+    1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/contains=] |element|:
+        1. Return false.
+    1. If |element| [=map/contains=] "{{SanitizerElementNamespaceWithAttributes/attributes}}":
+        1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+            1. [=map/iterate|For each=] |attr| in |configuration|["{{SanitizerConfig/attributes}}"]:
+                1. [=list/Remove=] |attr| from |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
+        1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=]:
+            1. [=map/iterate|For each=] |attr| in |element| [=map/contains=]
```
evilpie: Spurious `[=map/contains=]` ?

index.bs
```diff
@@ -541,74 +730,193 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+        1. Return false
+    1. Return true.
 
 </div>
 
 <div algorithm>
 To <dfn for="Sanitizer">replace an element with its children</dfn> |element| from a {{SanitizerConfig}} |configuration|, do:
 
 1. Set |element| to the result of [=canonicalize a sanitizer element=] with |element|.
+1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/contains=] |element:
```
evilpie: Missing `|` after `|element`

index.bs
```diff
@@ -541,74 +730,193 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
-    {{SanitizerConfig/removeElements}} and {{SanitizerConfig/replaceWithChildrenElements}} lists.
+1. [=SanitizerConfig/Remove=] |element| from
+    |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"].
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+    1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/contains=] |element|:
+        1. Return false.
+    1. If |element| [=map/contains=] "{{SanitizerElementNamespaceWithAttributes/attributes}}":
+        1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+            1. [=map/iterate|For each=] |attr| in |configuration|["{{SanitizerConfig/attributes}}"]:
+                1. [=list/Remove=] |attr| from |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
```
evilpie: Don't we need to mirror what we do for **_allow an attribute_** and also return false for these cases where "global" `attributes` already contains the attribute?

index.bs
```diff
@@ -541,74 +730,193 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+                |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
+            1. Set |result| to false.
+        1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
+           [=map/with default=] &laquo; [] &raquo; [=list/contains=] |attribute|:
+            1. [=list/Remove=] |attribute| from
+                |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"].
+            1. Set |result| to false.
+    1. Return |result|.
+1. Otherwise:
+    1. if |configuration|["{{SanitizerConfig/removeAttributes}}"] does not [=map/exist=]:
```
evilpie: Always false, remove.

index.bs
```diff
@@ -541,74 +730,193 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+                |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
+            1. Set |result| to false.
+        1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
+           [=map/with default=] &laquo; [] &raquo; [=list/contains=] |attribute|:
+            1. [=list/Remove=] |attribute| from
+                |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"].
+            1. Set |result| to false.
+    1. Return |result|.
+1. Otherwise:
+    1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] does not [=map/exist=], then
```
evilpie: Always false, remove.

index.bs
```diff
@@ -541,74 +730,193 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
 1. Set |configuration|["{{SanitizerConfig/comments}}"] to |allow|.
+1. Return true.
 
 </div>
 
 <div algorithm>
 To <dfn for="Sanitizer">set data attributes</dfn> with |allow| on a {{SanitizerConfig}} |configuration|, do:
 
+1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=], then return false.
+1. If |configuration|["{{SanitizerConfig/attributes}}"] does not [=map/exist=], then set
```
evilpie: Always false, remove.

index.bs
otherdaniel: Our use of global attributes matches this definition almost exactly: _"[Global attributes are attributes common to all HTML elements; they can be used on all elements, though they may have no effect on some elements.](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes)"_ The definition in ([[HTML]])[https://html.spec.whatwg.org/#global-attributes] isn't quite as nice for our purposes, but seems to hint at the same thing.

Would _"globally defined attributes"_ and _"per-element defined attributes"_ make sense?

index.bs

(That was my intuitition, but I don't think it has much of a basis.)

index.bs
otherdaniel: I'm quite sure `{}` ought to be valid, and that has neither `elements` nor `removeElements`.

index.bs
otherdaniel: I don't think this is true. On what ground would `{}` be invalid?

index.bs

index.bs

index.bs
otherdaniel: Yes, but I've put it a bit further up.

index.bs

index.bs

index.bs
otherdaniel: Yes. It shoul d be a subset.

index.bs

index.bs
otherdaniel: Not sure I get this? This section essentially implements a 4-way if, to implement our 2x2 decision table. This is for the case with both local and global attributes.

index.bs

index.bs

index.bs

index.bs
otherdaniel:
```js
s = new Sanitizer({});
s.allowAttribute("bla");
```

index.bs
otherdaniel: (as above)

index.bs
otherdaniel: (as above)

index.bs
otherdaniel: This shows up in a bunch of places. I think we need to agree on what happens with `{}`. If that always gets normalized to `{remove: [], removeAttributes:[]}` then we can indeed drop a number of these cases.

index.bs
otherdaniel: Not sure. My thinking was that the element part of 'allow an element' was successful, thus... Not sure this makes sense. Let's briefly discuss this...

index.bs
evilpie: I think we discussed this before. It's also part of your summary in https://github.com/WICG/sanitizer-api/issues/281#issuecomment-2910241555.

index.bs
```diff
@@ -468,23 +631,47 @@ template contents). It consistes of these steps:
...
-            [=Attr/namespace=] is `null` and
-            |configuration|["{{SanitizerConfig/dataAttributes}}"] is true
+      1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=] and
+          |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+          [=map/exists=]:
+          1. If |configuration|["{{SanitizerConfig/attributes}}"] does not
+              [=SanitizerConfig/contain=] |attrName| and
+              |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+              does not [=SanitizerConfig/contain=] |attrName|, and if
+              "data-" is not a [=code unit prefix=] of |attribute|'s [=Attr/local name=] aornd
```
evilpie:
```suggestion
              "data-" is not a [=code unit prefix=] of |attribute|'s [=Attr/local name=] and
```

index.bs
```diff
@@ -468,23 +631,47 @@ template contents). It consistes of these steps:
...
+              [=Attr/namespace=] is not `null` or
+              |configuration|["{{SanitizerConfig/dataAttributes}}"] is not true:
+              1. [=/Remove=] |attribute| from |child|.
```
evilpie:
```suggestion
              1. [=/remove an attribute|Remove=] |attribute| from |child|.
```

I am not sure if the "from |child|" part is necessary, but we need to use the right algorithm.

otherdaniel: Done. (Removed the `from |child|` bit. That algorithm finds the parent by itself.)

index.bs

index.bs
```diff
@@ -535,167 +738,319 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+
+The {{Sanitizer}} stores the [=Sanitizer/configuration=] in a canonical form, as this makes
+a number of processing steps easier.
+
+<div class=example>An {{SanitizerConfig/elements}} list `{elements: ["div"]}` gets stored as
+`{elements: [{name: "div", namespace: "http://www.w3.org/1999/xhtml"}]`).
+</div>
+
+<div algorithm>
+To <dfn for=Sanitizer>canonicalize the configuration</dfn> {{SanitizerConfig}} |configuration|
```
evilpie, line 1016: Depending on your definition of *canonical*, I would say the result is not necessarily in *canonical* form, e.g. has duplicates across different `elements` and `removeElements` etc.

evilpie:
Splitting the configuration initialization and verification in two parts actually seems like a good idea from a specification perspective. I do suspect that for implementation we would probably want to merge them again, but this should hopefully be unobservable.

index.bs
```diff
@@ -535,167 +738,319 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
-NOTE: [=SanitizerConfig/Remove=] matches on name and namespace, so adding an
-    element with attributes would still remove the matching element from the
-    {{SanitizerConfig/removeElements}} and {{SanitizerConfig/replaceWithChildrenElements}} lists.
+1. Set |element| to the result of [=canonicalize a sanitizer element with attributes=] with
+    |element|.
+1. [=SanitizerConfig/Remove=] |element| from
+    |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"].
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+    1. If |configuration|["{{SanitizerConfig/elements}}"] does not [=map/contain=] |element|:
+        1. [=Comment=]: This is the case with a global allow-list that does not yet contain
```
evilpie, line 773: This seems wrong to me. We always have to make sure that the attributes are valid.

evilpie: This seems wrong to me. We always have to make sure that the attributes are valid.

index.bs
```diff
@@ -535,167 +738,319 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] to the
+                [=set/difference=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] and
+                |configuration|["{{SanitizerConfig/attributes}}"].
+        1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=]:
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] to the
+                [=set/difference=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] and
+                |configuration|["{{SanitizerConfig/removeAttributes}}"].
+    1. If |element| equals |original element|, then return false.
```
evilpie: I think it's obvious what you are doing here, but we need to make sure this a defined operation.

index.bs
```diff
@@ -535,167 +738,319 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+
+1. If neither |configuration|["{{SanitizerConfig/elements}}"] nor
+    |configuration|["{{SanitizerConfig/removeElements}}"] [=map/exist=], then [=map/set=]
+    |configuration|["{{SanitizerConfig/removeElements}}"] to &laquo; [] &raquo;.
+1. If neither |configuration|["{{SanitizerConfig/attributes}}"] nor
+    |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exist=], then [=map/set=]
+    |configuration|["{{SanitizerConfig/removeAttributes}}"] to &laquo; [] &raquo;.
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+   1. [=list/iterate|For each=] |element| of |configuration|["{{SanitizerConfig/elements}}"] do:
+      1. Replace |element| with the result of [=canonicalize a sanitizer element with attributes=]
```
evilpie: I think we can't detect duplicate elements when doing it like this?

index.bs
```diff
@@ -535,167 +738,319 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+   1. [=list/iterate|For each=] |attribute| of |configuration|["{{SanitizerConfig/attributes}}"] do:
+      1. Replace |attribute| with the result of [=canonicalize a sanitizer attribute=]
+          |attribute|.
+1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=]:
+   1. [=list/iterate|For each=] |attribute| of |configuration|["{{SanitizerConfig/removeAttributes}}"] do:
+      1. Replace |attribute| with the result of [=canonicalize a sanitizer attribute=]
+          |attribute|.
+1. If |configuration|["{{SanitizerConfig/comments}}"] does not [=map/exist=], then [=map/set=]
+    |configuration|["{{SanitizerConfig/comments}}"] to |allowCommentsAndDataAttributes|.
+1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=] and
```
evilpie, line 1122: Shouldn't  `{ removeAttributes: [], dataAttributes: true }` be an error as well? I think we can't detect it, if we do it like this.

evilpie (changes requested):

index.bs
evilpie: We decided to make the exclusive, so please update this and the next step as well.

index.bs
```diff
@@ -368,6 +374,181 @@ dictionary SanitizerConfig {
...
+A {{SanitizerConfig}} |config| is <dfn for=SanitizerConfig>valid</dfn> if all of the following
+conditions hold:
+
+1. The |config| has an {{SanitizerConfig/elements}} or an {{SanitizerConfig/removeElements}}
+    [=map/key=], or neither, but not both.
+1. The |config| has an {{SanitizerConfig/attributes}} or an {{SanitizerConfig/removeAttributes}}
+    [=map/key=], or neither, but not both.
+1. Assuming that all element names have run through [=canonicalize a sanitizer element=] algorithm,
+    and all attribute names have run through [=canonicalize a sanitizer attribute=] algorithm.
+    1. The [=set/difference=] of |config|[{{SanitizerConfig/elements}}] and
```
evilpie: I think we need to handle `replaceWithChildrenElements` and `elements` being missing?

index.bs
evilpie: I believe this is still **unresolved**.

index.bs
```diff
@@ -368,6 +374,192 @@ dictionary SanitizerConfig {
...
+
+<div algorithm>
+A {{SanitizerConfig}} |config| is <dfn for=SanitizerConfig>valid</dfn> if all of the following
+conditions hold:
+
+1. The |config| has either an {{SanitizerConfig/elements}} or a {{SanitizerConfig/removeElements}}
+    [=map/key=], but not both.
+1. The |config| has either an {{SanitizerConfig/attributes}} or a {{SanitizerConfig/removeAttributes}}
+    [=map/key=], but not both.
+1. Assuming that all element names have run through [=canonicalize a sanitizer element=] algorithm,
```
evilpie: Can we change this into some kind of vague assert or something?

index.bs
```diff
@@ -368,6 +374,192 @@ dictionary SanitizerConfig {
...
+    and all attribute names have run through [=canonicalize a sanitizer attribute=] algorithm.
+    1. If both |config|[{{SanitizerConfig/elements}}] and
```
evilpie: I don't think we need this indentation.

index.bs
```diff
@@ -368,6 +374,192 @@ dictionary SanitizerConfig {
...
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exist=], then
+        the [=set/difference=] of |config|[{{SanitizerConfig/elements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
+    1. If both |config|[{{SanitizerConfig/removeElements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exist=], then
+        the [=set/difference=] of |config|[{{SanitizerConfig/removeElements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
+    1. If |config|[{{SanitizerConfig/attributes}}] [=map/exists=]:
+        1. [=list/iterate|For any=] |element| in |config|[{{SanitizerConfig/elements}}]:
```
evilpie, line 544: Need to check that `elements` exists

index.bs
```diff
@@ -368,6 +374,192 @@ dictionary SanitizerConfig {
...
+            1. |element|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}] is a
+                [=set/subset=] of |config|[{{SanitizerConfig/attributes}}].
+            1. If {{SanitizerConfig/dataAttributes}} is true:
+                1. |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] does not
+                    contain a [=custom data attribute=].
+        1. If {{SanitizerConfig/dataAttributes}} is true:
+            1. |config|[{{SanitizerConfig/attributes}}] does not contain
+                a [=custom data attribute=].
+    1. If |config|[{{SanitizerConfig/removeAttributes}}] [=map/exists=]:
+        1. [=list/iterate|For any=] |element| in |config|[{{SanitizerConfig/elements}}]:
```
evilpie, line 545: `elements` might not exist.

evilpie (changes requested):

index.bs
```diff
@@ -453,52 +645,75 @@ template contents). It consistes of these steps:
...
-    1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=SanitizerConfig/contains=] |elementName|, or if |configuration|["{{SanitizerConfig/elements}}"] is not [=list/empty=] and does not [=SanitizerConfig/contain=] |elementName|:
-       1. [=/Remove=] |child|.
-       1. [=Continue=].
-    1. If |elementName| [=equals=] &laquo;[ "`name`" &rightarrow; "`template`",
+    1. If |configuration|["{{SanitizerConfig/removeElements}}"] [=map/exists=] and
+        |configuration|["{{SanitizerConfig/removeElements}}"]  [=SanitizerConfig/contains=]
+        |elementName|:
+      1. [=/Remove=] |child|.
+      1. [=Continue=].
+    1. Otherwise:
```
evilpie: Sorry, maybe I made a mistake in my own suggestion, but this isn't quite right. We also get to this otherwise branch if `["removeElements"]` doesn't contain elementName.

Maybe like:

1. If configuration["removeElements"] exists:
   1. If configuration["removeElements"] contains elementName:
       1. Remove
       2. Continue
1. Otherwise,
    1. Assert: configuration["elements"] exists.
    2. ...

index.bs
```diff
@@ -368,6 +374,192 @@ dictionary SanitizerConfig {
...
+        the [=set/difference=] of |config|[{{SanitizerConfig/elements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
+    1. If both |config|[{{SanitizerConfig/removeElements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exist=], then
+        the [=set/difference=] of |config|[{{SanitizerConfig/removeElements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
+    1. If |config|[{{SanitizerConfig/attributes}}] [=map/exists=]:
+        1. [=list/iterate|For any=] |element| in |config|[{{SanitizerConfig/elements}}]:
+            1. The [=set/difference=] of |config|[{{SanitizerConfig/attributes}}] and
+                |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] is
```
evilpie: Actually we probably need a check for all of these element attributes/removeAttributes as well (including the next loop)

evilpie (changes requested):

index.bs
```diff
@@ -535,167 +750,331 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+        return false.
+    1. [=list/Append=] |attribute| to |configuration|["{{SanitizerConfig/attributes}}"]
+    1. [=Comment=]: Fix-up per-element allow and remove lists.
+    1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+        1. [=list/iterate|For each=] |element| in |configuration|["{{SanitizerConfig/elements}}"]:
+            1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+               [=map/with default=] &laquo; [] &raquo; [=list/contains=] |attribute|:
+                1. [=list/Remove=] |attribute| from
+                    |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
+            1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
```
evilpie: We can't have an attribute in the local removeAttributes list if it wasn't in the global attributes list. Maybe turn this into an assert?

index.bs
```diff
@@ -535,167 +750,331 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+    1. [=Comment=]: We need to make sure the per-element attributes do not overlap with global
+        attributes.
+    1. Let |original element| be a copy of |element|.
+    1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/exists=]:
+        1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] to the
+                [=set/difference=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] and
+                |configuration|["{{SanitizerConfig/attributes}}"].
+        1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=]:
```
evilpie, line 819: 1. Otherwise:

index.bs
```diff
@@ -535,167 +750,331 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+                [=set/intersection=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] and
+                |configuration|["{{SanitizerConfig/removeAttributes}}"].
+    1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] [=map/exists=]:
+        1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] to the
+                [=set/difference=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] and
+                |configuration|["{{SanitizerConfig/attributes}}"].
+        1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=]:
```
evilpie, line 831: Otherwise:

evilpie (changes requested):

index.bs
```diff
@@ -535,167 +750,331 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] to the
+                [=set/intersection=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] and
+                |configuration|["{{SanitizerConfig/removeAttributes}}"].
+    1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] [=map/exists=]:
+        1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] to the
+                [=set/difference=] of
```
evilpie: All local removeAttributes must be part of the global attributes, so this must an intersection.

index.bs
```diff
@@ -535,167 +750,331 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+    1. Let |original element| be a copy of |element|.
+    1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/exists=]:
+        1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] to the
+                [=set/difference=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] and
+                |configuration|["{{SanitizerConfig/attributes}}"].
+        1. If |configuration|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=]:
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] to the
+                [=set/intersection=] of
```
evilpie: Ah, this must be difference?

The only special case per your matrix is local remove, global allow. The rest is just no duplicates.

I believe the following would be correct:
```
var s = new Sanitizer({ elements: [], attributes: ["title", "dir"] });
s.allowElement({name: "b", attributes: ["dir", "lang"]});
console.log(s.get().elements[0].attributes); // "lang"

var s = new Sanitizer({ elements: [], removeAttributes: ["title", "dir"] });
s.allowElement({name: "b", attributes: ["dir", "lang"]});
console.log(s.get().elements[0].attributes); // "lang"

var s = new Sanitizer({ elements: [], attributes: ["title", "dir"] });
s.allowElement({name: "b", removeAttributes: ["dir", "lang"]})
console.log(s.get().elements[0].removeAttributes); // "dir"

var s = new Sanitizer({ elements: [], removeAttributes: ["title", "dir"] });
s.allowElement({name: "b", removeAttributes: ["dir", "lang"]});
console.log(s.get().elements[0].removeAttributes); // "lang"
```

index.bs
```diff
@@ -535,167 +750,331 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+                [=set/difference=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] and
+                |configuration|["{{SanitizerConfig/removeAttributes}}"].
+    1. If |configuration|["{{SanitizerConfig/elements}}"] does not [=map/contain=] |element|:
+        1. [=Comment=]: This is the case with a global allow-list that does not yet contain
+            |element|.
+        1. [=list/Append=] |element| to |configuration|["{{SanitizerConfig/elements}}"].
+        1. Return true.
+    1. [=Comment=]: This is the case with a global allow-list that already contains |element|.
+    1. If |element| [=map/equals=] |original element|, then return false.
```
evilpie: So, if the element had to modified we return early and don't overwrite the existing element in configuration["elements"] ? Or was that supposed to compare against the existing element?

otherdaniel: Yeah, I think it's meant to compare against the existing element.

What I wrote is whether the incoming |element| after normalization equals the original one. (That is, the normalization didn't affect |element|.)  But you're right, it should have said that the element post normalization doesn't affect the config (that is, we aren't actually making a change).

index.bs
```diff
@@ -368,6 +374,192 @@ dictionary SanitizerConfig {
...
+
+1. The |config| has either an {{SanitizerConfig/elements}} or a {{SanitizerConfig/removeElements}}
+    [=map/key=], but not both.
+1. The |config| has either an {{SanitizerConfig/attributes}} or a {{SanitizerConfig/removeAttributes}}
+    [=map/key=], but not both.
+1. Assuming that all element names have run through [=canonicalize a sanitizer element=] algorithm,
+    and all attribute names have run through [=canonicalize a sanitizer attribute=] algorithm.
+    1. If both |config|[{{SanitizerConfig/elements}}] and
+        |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exist=], then
+        the [=set/difference=] of |config|[{{SanitizerConfig/elements}}] and
```
evilpie: The *intersection* of `elements` and `replaceWithChildrenElements` must be empty.

index.bs
```diff
@@ -535,167 +750,331 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+* Whether we have a global allow- or remove-list,
+* whether they already contain |element| or not.
+
+</div>
+
 1. Set |element| to the result of [=canonicalize a sanitizer element=] with |element|.
-1. [=SanitizerConfig/Add=] |element| to |configuration|["{{SanitizerConfig/removeElements}}"].
-1. [=SanitizerConfig/Remove=] |element| from |configuration|["{{SanitizerConfig/elements}}"] list.
 1. [=SanitizerConfig/Remove=] |element| from |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"].
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
```
evilpie, line 879: While writing tests, I realized that removing the element from `replaceWithChildrenElements` obviously modifies the list, but we might still return false.

```javascript
let s = new Sanitizer({
    replaceWithChildrenElements: ["b"], 
    elements: []
});
s.removeElement("b"); // false
s.get()  //  { replaceWithChildrenElements: [], elements: [] }
```

evilpie (changes requested):

index.bs
evilpie: And the mirrored case of course:

```javascript
let s = new Sanitizer({
    replaceWithChildrenElements: ["b"], 
    removeElements: []
});
s.allowElement("b"); // false
s.get() // { replaceWithChildrenElements: [], removeElements: [] }
```

index.bs
evilpie: _Resolved_

index.bs
evilpie: _Resolved_

index.bs
evilpie: > 6.1.1.2. The difference of config[attributes] and element[attributes] is empty.
> 6.1.1.3. element[removeAttributes] is a subset of config[attributes].

> 7.1.1. The difference of config[removeAttributes] and element[attributes] is empty.
> 7.1.2. The difference of config[removeAttributes] and element[removeAttributes] is empty.

We still need to check those for existence.

evilpie:
Sorry, I tried to comment on some of the older comments and Github totally didn't like that.

index.bs
evilpie: **Resolved**

index.bs
evilpie: @otherdaniel I think this still an issue.

otherdaniel: Yes, that was still broken. I think trying to "optimize" the branches just makes it more complicated. It's now written so that:
- Each key has its own top-level if.
- If a node is removed, it continues the loop. That is, [=/remove=] and [=continue=] are always paired. If the node remains in the tree, then it'll execute the rest of the loop body.

Because we know that a valid config cannot have both elements and removeElements at the same time, there is arguably some redundancy in those if-statements, but I think trying to put that into the algorithm just makes it more complicated without really gaining anything.

index.bs
otherdaniel: Yes. Done, by defaulting them to `[]`.

index.bs
otherdaniel: We no longer us "implicit" return values.

index.bs
otherdaniel: This text no longer exists. Resolving the comment.

index.bs
otherdaniel: Updating of local lists has been added.

index.bs
otherdaniel: `{}` is now canonicalized to have empty `removeElements` and `removeAttribtues` keys, and is not indeed in invalid canonicalized config. Marking as resolved.

index.bs
otherdaniel: Canonicalization was changed to add keys to `{}`, resolving this issue.

index.bs
otherdaniel: The attribute handling now preceeds this bit of the algorithm. Marking this as resolved.

index.bs
otherdaniel: Added a [=map/equals=] helper definition.

index.bs
otherdaniel: Latest upload solves this: [=valid=] now checks for dupes with a [=has dupes=] helper algorithm, and canonicalization just adds anything it finds.

index.bs
otherdaniel: [=valid=] now treats this as an error.

index.bs
otherdaniel: Done. Also elsewhere in this algorithm.

index.bs

index.bs
otherdaniel: Done. (We should now consistently return false if any changed was made.)

index.bs
otherdaniel: Done.

index.bs
otherdaniel: Done.

index.bs
otherdaniel: Done. (Here & elsewhere. We now use a helper var |modified| to not "loose" the modification status.)

index.bs
otherdaniel: Removed. (Then later, removed that bit of text, but still... )

index.bs
otherdaniel: Done.

index.bs

index.bs
```diff
@@ -535,167 +767,354 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+        attributes.
+    1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/exists=]:
+        1. If |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] to the
+                [=set/difference=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] and
+                |configuration|["{{SanitizerConfig/attributes}}"].
+            1. If |configuration|["{{SanitizerConfig/dataAttributes}}"] [=map/exists=] and
+                |configuration|["{{SanitizerConfig/dataAttributes}}"] is true:
+                1. [=list/remove=] all items |item| from
```
evilpie: Nit: Capitalize _remove_.

index.bs
```diff
@@ -535,167 +767,354 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+    1. [=Comment=]: Fix-up per-element allow and remove lists.
+    1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+        1. [=list/iterate|For each=] |element| in |configuration|["{{SanitizerConfig/elements}}"]:
+            1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+               [=map/with default=] &laquo; [] &raquo; [=list/contains=] |attribute|:
+                1. [=list/Remove=] |attribute| from
+                    |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
+            1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
+               [=map/with default=] &laquo; [] &raquo; [=list/contains=] |attribute|:
+                1. [=Assert=]:
```
evilpie: I think this should just be: 
> 2. Assert: element["removeAttributes"] with default « [] » does not contain attribute.

Currently the assert is unreachable.

evilpie (approved):

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
 };
 </pre>
 
+## Configuration Invariants ## {#invariants}
+
+Configurations can and should be modified by developers to suit their
```
```suggestion
Configurations can and ought to be modified by developers to suit their
```
(should is a normative term)

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+modifier methods, or to {{Sanitizer/get()}} an existing {{Sanitizer}}'s
+[[#config|configuration]]  as a dictionary and modify the dictionary and then
+create a new {{Sanitizer}} with it.
+
+An empty configuration allows everything (when called with the "unsafe" methods like
+{{Element/setHTMLUnsafe}}).
+A configuration `"default"` contains a [=built-in safe default configuration=]. Note
+that "safe" and "unsafe" sanitizer methods have different defaults.
+
+Not all configuration dictionaries are valid. A valid configuration should avoid redundancy
```
```suggestion
Not all configuration dictionaries are valid. A valid configuration avoids redundancy
```

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+(like specifying the same element to be allowed twice) or contradictions (like specifying an
```
```suggestion
(like specifying the same element to be allowed twice) and contradictions (like specifying an
```

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+element to be both removed and allowed.)
+
+Several conditions need to hold for a configuration to be valid:
+
+- Mixing global allow- and remove-lists:
+  - {{SanitizerConfig/elements}} or {{SanitizerConfig/removeElements}} may exist,
```
```suggestion
  - {{SanitizerConfig/elements}} or {{SanitizerConfig/removeElements}} can exist,
```
(may is a normative term)

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+    but not both.
+    If both are missing, this is equivalent to {{SanitizerConfig/removeElements}}
+    [=map/set=] to &laquo; [] &raquo;.
+  - {{SanitizerConfig/attributes}} or {{SanitizerConfig/removeAttributes}} may exist,
```
```suggestion
  - {{SanitizerConfig/attributes}} or {{SanitizerConfig/removeAttributes}} can exist,
```

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+    but not both.
+    If both are missing, this is equivalent to {{SanitizerConfig/removeAttributes}}
+    [=map/set=] to &laquo; [] &raquo;.
+  - {{SanitizerConfig/dataAttributes}} is conceptually an extension of the
+    {{SanitizerConfig/attributes}} allow-list. The {{SanitizerConfig/dataAttributes}}
+    attribute is only allowed when a {{SanitizerConfig/attributes}} list is used.
+- Duplicate entries between different global lists:
+  - There are no duplicate entries (i.e. no same elements) between
```
```suggestion
  - There are no duplicate entries (i.e., no same elements) between
```

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+    {{SanitizerConfig/elements}}, {{SanitizerConfig/removeElements}}, or
+    {{SanitizerConfig/replaceWithChildrenElements}}.
+  - There are no duplicate entries (i.e. no same attributes) between
```
```suggestion
  - There are no duplicate entries (i.e., no same attributes) between
```

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+        listed in any of the allow-lists.
+  * If a global {{SanitizerConfig/removeAttributes}} remove list exists, then:
+    * If a local {{SanitizerElementNamespaceWithAttributes/attributes}} allow lists exists,
+        there may be no duplicate entries between these lists.
+    * If a local {{SanitizerElementNamespaceWithAttributes/removeAttributes}} remove lists exists,
+        there may be no duplicate entries between these lists.
+    * {{SanitizerConfig/dataAttributes}} must be absent.
+
+
+Note that these rules put a somewhat higher burden on configuration authors, as it's on to provide
```
```suggestion
Note that these rules put a somewhat higher burden on configuration authors, as it's on them to provide
```

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+conditions hold:
+
+1. The |config| has either an {{SanitizerConfig/elements}} or a {{SanitizerConfig/removeElements}}
+    [=map/key=], but not both.
+1. The |config| has either an {{SanitizerConfig/attributes}} or a {{SanitizerConfig/removeAttributes}}
+    [=map/key=], but not both.
+1. [=Assert=]: All {{SanitizerElementNamespaceWithAttributes}}, {{SanitizerElementNamespace}}, and
+    {{SanitizerAttributeNamespace}} items in |config| are canonical, meaning they have been run
+    through [=canonicalize a sanitizer element=] or [=canonicalize a sanitizer attribute=],
+    as appropriate.
```
annevk, line 528: This will need some expanding, but it's probably okay for the initial draft.

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+1. None of |config|[{{SanitizerConfig/elements}}], |config|[{{SanitizerConfig/removeElements}}],
+    |config|[{{SanitizerConfig/replaceWithChildrenElements}}],
+    |config|[{{SanitizerConfig/attributes}}], or
+    |config|[{{SanitizerConfig/removeAttributes}}], if they [=map/exist=],
+    [=SanitizerConfig/has dupes=].
+1. If both |config|[{{SanitizerConfig/elements}}] and
+    |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exist=], then
+    the [=set/intersection=] of |config|[{{SanitizerConfig/elements}}] and
+    |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
```
annevk, line 537: I don't think this works well since elements contains more complicated entries. You first have to transform them into a simple set without the attributes.

index.bs
```diff
@@ -368,6 +378,207 @@ dictionary SanitizerConfig {
...
+  * If a global {{SanitizerConfig/removeAttributes}} remove list exists, then:
+    * If a local {{SanitizerElementNamespaceWithAttributes/attributes}} allow lists exists,
+        there may be no duplicate entries between these lists.
+    * If a local {{SanitizerElementNamespaceWithAttributes/removeAttributes}} remove lists exists,
+        there may be no duplicate entries between these lists.
+    * {{SanitizerConfig/dataAttributes}} must be absent.
+
+
+Note that these rules put a somewhat higher burden on configuration authors, as it's on to provide
+a configuration that is valid.
```
annevk: Though this is not entirely accurate because `{}` is invalid, but is a fine input. And once you have a `Sanitizer` object it's always valid. So maybe we shouldn't include this.

index.bs
```diff
@@ -535,167 +767,354 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
-## Configuration Processing ## {#configuration-processing}
+## Modify the Configuration ## {#configuration-modify}
+
+The configuration modifier methods are methods on {{Sanitizer}} that modify its configuration.
+They will maintain the validity criteria.
+They return a boolean which informs the caller whether the configuration was modified or not.
+
+<div class=example>
+* `new Sanitizer({elements: ["div"]}).allowElement("p");` &mdash; Returns true. Allows "div" and "p".
+* `new Sanitizer({elements: ["div"]}).removeElement("p");` &mdash; Returns false. Allows "div".
```
annevk: I think you should flush this out a bit because if you don't assign the instance to a variable it's entirely meaningless and doesn't really allow anything.

index.bs
```diff
@@ -535,167 +767,354 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+
+</div>
 
 <div algorithm>
-To <dfn for="SanitizerConfig">allow an element</dfn> |element| with a {{SanitizerConfig}} |configuration|, do:
+To <dfn for="SanitizerConfig">allow an element</dfn> |element| with a
```
annevk: What's the type of element here?

index.bs
```diff
@@ -535,167 +767,354 @@ To determine whether an |attribute| <dfn>contains a javascript: URL</dfn>:
...
+{{SanitizerConfig}} |configuration|, do:
```
```suggestion
{{SanitizerConfig}} |configuration|:
```

annevk (approved):
By-and-large this looks okay to me, but there's many editorial bits to follow-up on.

index.bs

index.bs

index.bs

index.bs

index.bs

index.bs
otherdaniel: Done. 

Introduced [=SanitizerConfig/intersection=], which normalizes before calling [=set/intersection=].

index.bs

Seperated out the examples, and made them 3 instructions, rather than cramming the example into one.

index.bs
otherdaniel: Done. (Here & elsewhere.)

index.bs
otherdaniel: Done. (Here & elsewhere.)

index.bs

index.bs

index.bs
otherdaniel: I removed the whole paragraph.

index.bs
otherdaniel: Obsolete (since paragraph was removed.)

index.bs

index.bs

PR #298, opened 2025-06-12
Initial draft for SVG in default configuration.

otherdaniel:
Note: This hasn't been reviewed yet.

### DIFF
--- a/builtins/safe-default-configuration.json
+++ b/builtins/safe-default-configuration.json
@@ -582,14 +582,6 @@
       "namespace": "http://www.w3.org/1998/Math/MathML",
       "attributes": [
         {
-          "name": "width",
-          "namespace": null
-        },
-        {
-          "name": "height",
-          "namespace": null
-        },
-        {
           "name": "depth",
           "namespace": null
         },
@@ -632,14 +624,6 @@
       "name": "mspace",
       "namespace": "http://www.w3.org/1998/Math/MathML",
       "attributes": [
-        {
-          "name": "width",
-          "namespace": null
-        },
-        {
-          "name": "height",
-          "namespace": null
-        },
         {
           "name": "depth",
           "namespace": null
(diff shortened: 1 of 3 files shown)


builtins/safe-default-configuration.txt
```diff
@@ -243,3 +243,200 @@ math semantics
...
+- y
+- dx
+- dy
+- rotate
+- textLength
+svg textPath
+- lengthAdjust
+- textLength
+- path
+- href
```
annevk: This should probably be disallowed?

builtins/safe-default-configuration.txt
```diff
@@ -243,3 +243,200 @@ math semantics
...
+- startOffset
+- method
+- spacing
+- side
+
+// Embedded Content, https://svgwg.org/svg2-draft/embedded.html
+svg image
+- preserveAspectRatio
+- href
```
annevk: Ditto.

annevk (commented):
Nice work!

builtins/safe-default-configuration.txt
mozfreddyb: I think we don't disallow `href` per se but instead give them a [special pass for `javascript` URLs etc](https://wicg.github.io/sanitizer-api/#sanitization)

annevk: That's for `href` when it performs navigation. But SVG uses it for embedding.

builtins/safe-default-configuration.txt
```diff
@@ -243,3 +243,200 @@ math semantics
...
+- markerUnits
+- markerWidth
+- markerHeight
+- orient
+
+// Purposely omitted: svg script
+
+// Common attributes: https://svgwg.org/svg2-draft/struct.html#CommonAttributes
+[SVG Global]
+- id
```
evilpie: Observation: This means we now allow `id` for all elements (especially HTML).

builtins/safe-default-configuration.txt
```diff
@@ -243,3 +243,200 @@ math semantics
...
+- lang
+- xml space
+- tabindex
+- autofocus
```
evilpie: I don't think we want this.

builtins/safe-default-configuration.txt
```diff
@@ -243,3 +243,200 @@ math semantics
...
+- orient
+
+// Purposely omitted: svg script
+
+// Common attributes: https://svgwg.org/svg2-draft/struct.html#CommonAttributes
+[SVG Global]
+- id
+- lang
+- xml space
+- tabindex
```
evilpie: Actual a bit similar to autofocus, I think this could change the global tab order, which might make it easy to trick keyboard users into activating something they didn't mean to.

evilpie:
Thanks Daniel for doing this work and giving us something concrete to discuss.

builtins/safe-default-configuration.txt
evilpie: Couldn't this be used to trick code that looks up elements by their id (e.g. `getElementById`) to operate on the wrong element?

How important are ids in SVG? I know it's used for references in `<use>`, but we don't allow that.

builtins/safe-default-configuration.txt
otherdaniel: Done.

builtins/safe-default-configuration.txt
otherdaniel: Done.

builtins/safe-default-configuration.txt
otherdaniel: Done (that is, removed the id attribute.

builtins/safe-default-configuration.txt
otherdaniel: Removed.

builtins/safe-default-configuration.txt
otherdaniel: Done.

builtins/safe-default-configuration.txt
```diff
@@ -243,3 +243,211 @@ math semantics
...
+- textLength
+- path
+// - href, purposely omitted: This loads a textPath from a resource.
+- startOffset
+- method
+- spacing
+- side
+
+// Embedded Content, https://svgwg.org/svg2-draft/embedded.html
+svg image
```
otherdaniel: Not sure if `<svg:image>` without `href` attribute still makes sense...

evilpie: I don't think so. We also don't have ``, so I don't think we should have this either.

evilpie:
I tried scraping all content attributes from the HTML specific and I found 146 element specific attributes. I am not sure how close this is to the actual number.

Anyway, I found one irrelevant overlap with the SVG global attributes: The link element's [`color`](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-color) attribute. `<link>` is not in the default configuration.

builtins/safe-default-configuration.txt
otherdaniel: Removed.

builtins/safe-default-configuration.json
```diff
@@ -728,6 +728,416 @@
...
+          "namespace": null
+        },
+        {
+          "name": "y",
+          "namespace": null
+        }
+      ]
+    },
+    {
+      "name": "switch",
```
annevk: I'm wondering if we should leave this out for now, given its capability to allow displaying one type of content to user A and another to user B.

builtins/safe-default-configuration.json
```diff
@@ -762,6 +1172,258 @@
     {
       "name": "scriptlevel",
       "namespace": null
+    },
+    {
+      "name": "space",
+      "namespace": "http://www.w3.org/XML/1998/namespace"
```
annevk: If this is the only namespaced attribute we end up allowing let's scrap it as it's useless anyway.

annevk (commented):
@mozfreddyb we still need to decide whether to flatten/sink SVG global attributes into allowed SVG elements. I'm personally on the fence.

builtins/safe-default-configuration.json
evilpie: Yeah, I am on the fence about this as well. I actually think the removal of ARIA gives us precedence here.

(This means we can also remove `systemLanguage`)

builtins/safe-default-configuration.json
evilpie: It's marked as deprecated on [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/xml:space) as well.

mozfreddyb: 👍,  as above :)

builtins/safe-default-configuration.json
mozfreddyb: Good catch. I agree.

otherdaniel: Done.

builtins/safe-default-configuration.json
otherdaniel: Done.

builtins/safe-default-configuration.txt
```diff
@@ -243,3 +243,211 @@ math semantics
...
+[SVG Global]
+// - id, purposely omitted, as it interferes with HTML, MathML, and "DOM/id clobbering"
+- lang
+// - xml space, purposely omitted; deprecated and useless
+// - tabindex, purposely omitted, as it potentially interferes with other parts on the page
+// - autofocus, purposely omitted, as it potentially interferes with other parts on the page
+// data-*, purposely omitted because we have an attribute for that.
+// Conditional processing attributes, https://svgwg.org/svg2-draft/struct.html#ConditionalProcessingDefinitions
+- requiredExtensions
+- systemLanguage
```
evilpie: I think without `<switch>` both `requiredExtensions` and `systemLanguage` should also be removed.

otherdaniel: Will remove. (As I read this, the spec allows them on many elements, but those attributes create switch-like behaviour.)

otherdaniel: Done.

builtins/safe-default-configuration.json
```diff
@@ -728,6 +728,411 @@
...
+      "namespace": "http://www.w3.org/2000/svg",
+      "attributes": []
+    },
+    {
+      "name": "defs",
+      "namespace": "http://www.w3.org/2000/svg",
+      "attributes": []
+    },
+    {
+      "name": "symbol",
```
longsonr, line 773: I don't see any point in allowing symbol if you don't allow use.

builtins/safe-default-configuration.json
```diff
@@ -762,6 +1167,246 @@
...
+    {
+      "name": "color",
+      "namespace": null
+    },
+    {
+      "name": "color-interpolation",
+      "namespace": null
+    },
+    {
+      "name": "color-interpolation-filters",
```
longsonr, line 1204: if filters aren't allowed then color-interpolation-filters would seem non-functional.

builtins/safe-default-configuration.json
```diff
@@ -762,6 +1167,246 @@
...
+    {
+      "name": "fill-rule",
+      "namespace": null
+    },
+    {
+      "name": "filter",
+      "namespace": null
+    },
+    {
+      "name": "flood-color",
```
longsonr, line 1236: flood-color, flood-opacity are only used by the feFlood filter and that's not allowed is it?

builtins/safe-default-configuration.json
```diff
@@ -762,6 +1167,246 @@
...
+    {
+      "name": "font-variant",
+      "namespace": null
+    },
+    {
+      "name": "font-weight",
+      "namespace": null
+    },
+    {
+      "name": "glyph-orientation-horizontal",
```
longsonr, line 1272: glyph-orientation-horizontal and glyph-orientation-vertical are deprecated.

builtins/safe-default-configuration.json
```diff
@@ -762,6 +1167,246 @@
...
+      "namespace": null
+    },
+    {
+      "name": "glyph-orientation-vertical",
+      "namespace": null
+    },
+    {
+      "name": "image-rendering",
```
longsonr, line 1280: there's no img/image so image-rendering doesn't do anything.

builtins/safe-default-configuration.json
```diff
@@ -762,6 +1167,246 @@
...
+      "namespace": null
+    },
+    {
+      "name": "letter-spacing",
+      "namespace": null
+    },
+    {
+      "name": "lighting-color",
```
longsonr, line 1288: no filters no no need for lighting-color

builtins/safe-default-configuration.json
```diff
@@ -762,6 +1167,246 @@
...
+    {
+      "name": "marker-mid",
+      "namespace": null
+    },
+    {
+      "name": "marker-start",
+      "namespace": null
+    },
+    {
+      "name": "mask",
```
longsonr, line 1304: Are you allowing masks?

PR #299, opened 2025-07-03
De-normalize built-in safe default configuration, for better readability.

otherdaniel:
Configuration processing includes normalization steps, so that e.g. `"div"` gets turned into `{ "name": "div", "namespace": "http://www.w3.org/1999/xhtml" }`. The safe default config we include in the spec is already normalized.

The normalized configuration is rather verbose, and IMHO difficult to read. This patch modifies the script to "de-normlize" config text. This reduces the text by about 2/3.

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/299.html" title="Last updated on Jul 3, 2025, 12:11 PM UTC (f9a53c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/299/7f24d56...otherdaniel:f9a53c1.html" title="Last updated on Jul 3, 2025, 12:11 PM UTC (f9a53c1)">Diff</a>

### DIFF
--- a/builtins/safe-default-configuration.json
+++ b/builtins/safe-default-configuration.json
@@ -1,1413 +1,560 @@
 {
   "elements": [
-    {
-      "name": "html",
-      "namespace": "http://www.w3.org/1999/xhtml",
-      "attributes": []
-    },
-    {
-      "name": "head",
-      "namespace": "http://www.w3.org/1999/xhtml",
-      "attributes": []
-    },
-    {
-      "name": "title",
-      "namespace": "http://www.w3.org/1999/xhtml",
-      "attributes": []
-    },
-    {
-      "name": "body",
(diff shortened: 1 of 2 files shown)


otherdaniel:
Not sure this is useful. The intent is to make the "builtins" chapter easier to read. wdyt?

annevk:
I think in the end we want an English description rather than JSON, or at the very least we want to use Infra data types, with perhaps a link to the JSON file from which the English description is generated. And at that point I suspect it makes more sense if the JSON represents the normalized version, but it might make sense to offer a de-normalized version as well for those that prefer it? Though a bit unclear how many of these we want to maintain on an ongoing basis.

otherdaniel:
Closing, as on second view we weren't so sure this improves readability enough to be worth it.

Issue #290, opened 2025-05-14
Add SVG to the default allow list


annevk:
We wanted to track SVG separately from #103 as that contained a lot of MathML-related discussion and MathML is now included and tested.

evilpie:
I asked the SVG WG for help: https://github.com/w3c/svgwg/issues/981

evilpie:
See https://github.com/WICG/sanitizer-api/pull/298 for a draft suggestion.

annevk:
While #298 has now landed, there was a bunch of feedback left by @longsonr after landing that's worth evaluating: https://github.com/WICG/sanitizer-api/pull/298#pullrequestreview-2983166600

I'm okay with removing all the additional things they identified.

evilpie:
I am going to open a PR with his feedback.

mozfreddyb:
https://github.com/WICG/sanitizer-api/pull/300

PR #300, opened 2025-07-04
Address the SVG default configuration feedback

evilpie:
Thanks to @longsonr for providing feedback on #298.

Part of #290


builtins/safe-default-configuration.txt
```diff
@@ -397,33 +389,24 @@ svg marker
 - clip-rule
 - color
 - color-interpolation
-- color-interpolation-filters
```
annevk, line 400: Shouldn't we list these as purposely omitted as well? We do it for overflow below.

evilpie: Good idea.

annevk (approved):

otherdaniel (approved):

builtins/safe-default-configuration.json
```diff
@@ -1232,14 +1190,6 @@
       "name": "filter",
```
longsonr: filter can go too can't it?

evilpie: I think you are right, updating.

PR #309, opened 2025-09-18
Handle duplicate local attributes

### LINKED ISSUE #306: Handle duplicate attributes in local attributes/removeAttributes lists
I don't think we handle duplicate attributes inside the `elements`' local `attributes` or `removeAttributes` list anywhere. Both when constructing the Sanitizer (so in https://wicg.github.io/sanitizer-api/#sanitizerconfig-valid) or in [allow an element](https://wicg.github.io/sanitizer-api/#sanitizerconfig-allow-an-element).

evilpie:
This does the following:
- When canonicalizing SanitizerElementWithAttributes we don't remove duplicates in the local attributes/removeAttributes anymore.
- In the Sanitizer constructor we now throw for duplicates.
- In allowElement we now remove all duplicate attributes.

Fixes #306.

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/309.html" title="Last updated on Sep 22, 2025, 10:59 AM UTC (ed2fa93)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/309/51f78b0...evilpie:ed2fa93.html" title="Last updated on Sep 22, 2025, 10:59 AM UTC (ed2fa93)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -530,7 +530,7 @@ conditions hold:
     |config|[{{SanitizerConfig/replaceWithChildrenElements}}],
     |config|[{{SanitizerConfig/attributes}}], or
     |config|[{{SanitizerConfig/removeAttributes}}], if they [=map/exist=],
-    [=SanitizerConfig/has dupes=].
+    [=SanitizerConfig/has duplicates=].
 1. If both |config|[{{SanitizerConfig/elements}}] and
     |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exist=], then
     the [=SanitizerConfig/intersection=] of |config|[{{SanitizerConfig/elements}}] and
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -542,6 +542,8 @@ conditions hold:
 1. If |config|[{{SanitizerConfig/attributes}}] [=map/exists=]:
     1. If |config|[{{SanitizerConfig/elements}}] [=map/exists=]:
         1. [=list/iterate|For any=] |element| in |config|[{{SanitizerConfig/elements}}]:
+            1. Neither |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] or
+               |element|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}], if they exist, has dupes.
```
annevk: It should be defined what this means.

index.bs
```diff
@@ -1222,6 +1228,15 @@ there is more than one |entry| in |list| where
 |item|["namespace"] [=string/is|equals=] |entry|["namespace"].
 </div>
 
+<div algorithm>
+To <dfn for=SanitizerConfig>remove dupes</dfn> from a [=/list=] |list|,
```
annevk: duplicates*, no need for an abbreviation.

index.bs
```diff
@@ -1221,15 +1221,15 @@ of |A|'s [=map/entries=] and the [=ordered set=] of |B|'s [=map/entries=] are [=
 </div>
 
 <div algorithm>
-A [=/list=] |list| <dfn for=SanitizerConfig>has dupes</dfn>,
+A [=/list=] |list| <dfn for=SanitizerConfig>has duplicates</dfn>,
 if [=list/iterate|for any=] |item| of |list|,
 there is more than one |entry| in |list| where
 |item|["name"] [=string/is|equals=] |entry|["name"] and
 |item|["namespace"] [=string/is|equals=] |entry|["namespace"].
```
annevk: This definition doesn't work for attributes I'd think?

evilpie: Does it not work because the namespace could be null? Otherwise I am not quite sure what you mean, as attributes also have a name and namespace.

annevk: Ah right. So this is mostly correct except that we shouldn't use string/is. We should just write "is" without reference so it works for both strings and null.

PR #311, opened 2025-09-22
Don't allow local attributes+removeAttributes together with global removeAttributes

### LINKED ISSUE #305: Making the local attributes+removeAttributes with global removeAttributes special case unambigious
For the special case of a local `attributes` list combined with a global `removeAttributes` lists, we should probably just drop or throw for a local removeAttributes list.

```javascript
{
  removeAttributes: []
  elements: [{ name: "p", attributes: ["id"], removeAttributes: ["title"] }]
}
```
is the same as (because only attributes in the `attributes` list would be allowed)
```javascript
{
  removeAttributes: []
  elements: [{ name: "p", attributes: ["id"] }]
}
```

Even if we don't want to throw for this case in the `Sanitizer` constructor for some reason, we might want to think about doing some canonicalization in `allowElement()`.

evilpie:
- Add a new invariant, that both local `attributes` and `removeAttributes` are forbidden with a global `removeAttributes` list.
- Start throwing in the constructor
- Switches the conditions in allowElement (please double check..)
- In allowElement with a global removeAttributes lists, first make sure to remove everything in the local `removeAttributes` list from the local `attributes` list. Afterwards get rid of the disallowed local `removeAttributes` lists.

Fixes #305

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/311.html" title="Last updated on Oct 17, 2025, 7:13 AM UTC (e26b216)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/311/fc36ba1...evilpie:e26b216.html" title="Last updated on Oct 17, 2025, 7:13 AM UTC (e26b216)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -448,6 +448,16 @@ Several conditions need to hold for a configuration to be valid:
     {{SanitizerConfig/replaceWithChildrenElements}}.
   - There are no duplicate entries (i.e., no same attributes) between
     {{SanitizerConfig/attributes}} or {{SanitizerConfig/removeAttributes}}.
+- Mixing local allow- and remove-lists on the same element:
+  - When a {{SanitizerConfig/attributes}} list exists, both, either or none of the
+    {{SanitizerElementNamespaceWithAttributes/attributes}} and
+    {{SanitizerElementNamespaceWithAttributes/removeAttributes}}
+    lists are allowed  on the same element.
+  - When a {{SanitizerConfig/removeAttributes}} list exists, either or none of the
+    {{SanitizerElementNamespaceWithAttributes/attributes}} and
+    {{SanitizerElementNamespaceWithAttributes/removeAttributes}}
+    lists are allowed on the same element,
+    but not both.
 - Duplicate entries on the same element:
   - There are no duplicate entries between {{SanitizerElementNamespaceWithAttributes/attributes}}
     and {{SanitizerElementNamespaceWithAttributes/removeAttributes}} on the same element.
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -558,6 +570,8 @@ conditions hold:
 1. If |config|[{{SanitizerConfig/removeAttributes}}] [=map/exists=]:
     1. If |config|[{{SanitizerConfig/elements}}] [=map/exists=],
        then [=list/iterate|for any=] |element| in |config|[{{SanitizerConfig/elements}}]:
+        1. Not both |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] and
+            |element|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}] [=map/exist=].
```
annevk, line 608: I wonder if we should allow this, but only ban a combination of local attributes existing and removeAttributes being non-empty.

index.bs
```diff
@@ -816,20 +831,27 @@ allow- or remove-lists for attributes. This requires that we distinguish 4 cases
...
-            1. Set |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] to the
+
+    1. Otherwise:
+      1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/exists=]:
+        1. Set |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] to the
+                [=set/difference=] of
+                |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] and
+                |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
+                [=map/with default=] &laquo; [] &raquo;.
+        1. [=map/Remove=] |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"].
```
annevk, line 884: We also could empty it and keep it. But maybe removing it is slightly nicer for serialization purposes even if it means the same.

evilpie: We definitely want to remove the property if we keep the strict behavior in the constructor.

index.bs
evilpie: Not sure. I think it's maybe easier to explain that either of these properties are allowed, instead of attributes and an empty removeAttributes list are allowed, while the opposite is not true.

This does match the behavior for elements/removeElements, but that case isn't directly comparable of course, because `removeElements: []` actually means something.

So far, I feel like we have come down on the side of making the constructor stricter.

index.bs
annevk: I suppose I would allow the empty list in the constructor too as it's harmless.

I think I'm okay with the "either" alternative, but if your code gets a handle on a canonicalized element that initially didn't have these lists, you'd now have to actively remove the empty `removeAttributes` list if you wanted to add an `attributes` list.

evilpie:
@otherdaniel @mozfreddyb It would be good to get your opinion on this. This might be one of the last remaining points of contention.

index.bs
otherdaniel: I'd agree with Tom. So far, the spec text is consistent that if a list is present, then it can also be used. If we introduce a list that can only be empty (provided other conditions hold), this feels odd.

(Admittedly, this is of little consequence in this particular case since there's no API to seperately modify the per-element allow/remove lists. Still feels odd.)

otherdaniel:
> @otherdaniel @mozfreddyb It would be good to get your opinion on this. This might be one of the last remaining points of contention.

This looks good to me!

index.bs
evilpie: As discussed in the meeting. We will not allow both to exist for now. We can also be less strict in the future.

PR #312, opened 2025-09-22
Do nothing for allowElement with local attributes and a global removeElements list

### LINKED ISSUE #304: What should we do for `allowElement({ .. attributes .. })` with a `removeElements` list
With the changes being introduced by #296, we are only going to have either a `elements` or `removeElements` list. Modifier methods like `allowElement` modify the currently existing one.

I am wondering what we should do about cases like this:
```
let s = new Sanitizer({ removeElements: ["a", "p"] })
s.allowElement({ name: "a", removeAttributes: ["href"] }) 
```

With the current proposed behavior in #296, we would simply delete the `"a"` element from the `removeElements` list, but completely drop `removeAttributes: ["href"]`. Of course we don't have local attributes lists for anything but `elements`, so there isn't really anything we can do. But it does seem dangerous to me to just silently drop it (e.g. the link is unexpectedly clickable). Should we do something about this? I know we tried not to throw for any of the modifier methods. Should we at least have some suggestion about producing a console warning? Or maybe something else.

evilpie:
Local attributes/removeAttributes list are only supported with a global elements list. Otherwise we should just ignore the unsatisfiable request. I also added an optional warning, but we can remove that if it is unwanted.

Fixes #304

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/312.html" title="Last updated on Sep 22, 2025, 10:37 AM UTC (e54c6fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/312/51f78b0...evilpie:e54c6fa.html" title="Last updated on Sep 22, 2025, 10:37 AM UTC (e54c6fa)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -31,6 +31,7 @@ text: parse HTML from a string; type: dfn; url: https://html.spec.whatwg.org/#pa
 text: HTML fragment parsing algorithm; type: dfn; url: https://html.spec.whatwg.org/#html-fragment-parsing-algorithm
 text: global attribute; type: dfn; url: https://html.spec.whatwg.org/multipage/dom.html#global-attributes
 text: custom data attribute; type: dfn; url: https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute
+text: report a warning to the console; type: dfn; url: https://console.spec.whatwg.org/#report-a-warning-to-the-console
 </pre>
 <pre class="biblio">
 {
@@ -800,9 +801,9 @@ allow- or remove-lists for attributes. This requires that we distinguish 4 cases
 
 1. Set |element| to the result of [=canonicalize a sanitizer element with attributes=] with
     |element|.
-1. Set |modified| to the result of [=SanitizerConfig/remove=] |element| from
-    |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"].
 1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+    1. Set |modified| to the result of [=SanitizerConfig/remove=] |element| from
+       |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"].
     1. [=Comment=]: We need to make sure the per-element attributes do not overlap with global
         attributes.
     1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/exists=]:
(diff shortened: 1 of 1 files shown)


mozfreddyb (approved):

index.bs
```diff
@@ -849,8 +850,11 @@ allow- or remove-lists for attributes. This requires that we distinguish 4 cases
     1. [=list/Append=] |element| to |configuration|["{{SanitizerConfig/elements}}"]
     1. Return true.
 1. Otherwise:
-  1. [=Comment=]: If we have a global remove-list, the per-element attributes of |element|
-      get ignored.
+  1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] or |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] [=map/exists=]:
+    1. The user agent may [=report a warning to the console=] that this operation is not supported.
+    1. Return false.
```
annevk, line 855: How is this early return accurate? Don't we still want to remove the element from the various lists?

index.bs
```diff
@@ -849,8 +850,11 @@ allow- or remove-lists for attributes. This requires that we distinguish 4 cases
...
+  1. Set |modified| to the result of [=SanitizerConfig/remove=] |element| from
```
annevk, line 856: This should say "removing" for proper English, I think.

index.bs
evilpie: We don't want to remove it from `removeElements`, because that would make it allowed. I don't think we should be removing it from `replaceWithChildrenElements` either, because that would basically make it allowed as well, if it was previously in that list. I think I am most comfortable with not changing any of the lists and just totally ignoring the function call.

index.bs
evilpie: To be fair, I just copied this :)

index.bs
annevk: But isn't this the `allowElement()` method call?

I guess your point is that if you want to narrowly allow an element it can't work? But then I think we should at least allow for `removeAttributes` being the empty list, since that is what we canonicalize toward.

evilpie: Allowing empty `removeAttributes` makes sense to me. I will open a followup.

PR #316, opened 2025-09-23
Introduce a canonical form for SanitizerElementNamespaceWithAttributes

### LINKED ISSUE #304: What should we do for `allowElement({ .. attributes .. })` with a `removeElements` list

### LINKED ISSUE #308: Canonicalize missing or empty element's attributes/removeAttributes lists
I think [**canonicalize a sanitizer element with attributes**](https://wicg.github.io/sanitizer-api/#canonicalize-a-sanitizer-element-with-attributes) currently doesn't handle the absence of the `attributes` or `removeAttributes` lists at all and just assumes they are always present. That makes it difficult to tell what should happen in that case. So should `{ name: "div", attributes: [], removeAttributes: [] }` and `{ name: "div" }` (or even just `"div"`)  both result in the same dictionary? The difference is observable via `get()` and the [**equals**](https://wicg.github.io/sanitizer-api/#map-equal) check in [allow an element](https://wicg.github.io/sanitizer-api/#sanitizerconfig-allow-an-element).

evilpie:
- SanitizerElementNamespaceWithAttributes now always have at least an empty `removeAttributes` list. (Fixes #308)
- We handle missing properties correctly during **canonicalize a sanitizer element with attributes**
- Make sure to accept the canonical form (`{ name: .., namespace: ..., removeAttributes: [] }`)  when doing `allowElement()` with a `removeElements` lists. (Fixes #304)

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/316.html" title="Last updated on Sep 25, 2025, 3:09 PM UTC (2ce4551)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/316/b799894...evilpie:2ce4551.html" title="Last updated on Sep 25, 2025, 3:09 PM UTC (2ce4551)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -855,7 +855,9 @@ allow- or remove-lists for attributes. This requires that we distinguish 4 cases
     1. [=list/Append=] |element| to |configuration|["{{SanitizerConfig/elements}}"]
     1. Return true.
 1. Otherwise:
-  1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] or |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] [=map/exists=]:
+  1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/exists=] or
+     |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] [=with default=] &laquo; &raquo;
+     is not [=list/empty=]:
     1. The user agent may [=report a warning to the console=] that this operation is not supported.
     1. Return false.
   1. Set |modified| to the result of [=SanitizerConfig/remove=] |element| from


index.bs
```diff
@@ -1139,12 +1141,17 @@ To <dfn>canonicalize a sanitizer element with attributes</dfn> a {{SanitizerElem
...
 1. If |element| is a [=dictionary=]:
-   1. [=list/iterate|For each=] |attribute| in
-      |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]:
-      1. [=list/Append=] the result of [=canonicalize a sanitizer attribute=] with |attribute| to |result|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
-   1. [=list/iterate|For each=] |attribute| in
-      |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]:
-      1. [=list/Append=] the result of [=canonicalize a sanitizer attribute=] with |attribute| to |result|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"].
+   1. If neither |element|["{{SanitizerConfig/attributes}}"] nor
+      |element|["{{SanitizerConfig/removeAttributes}}"] [=map/exist=], then [=map/set=]
+      |result|["{{SanitizerConfig/removeAttributes}}"] to &laquo; [] &raquo;.
```
```suggestion
      |result|["{{SanitizerConfig/removeAttributes}}"] to &laquo; &raquo;.
```
Is what we want here I think. Set it to the empty list right? I saw this before so maybe this is wrong elsewhere as well.

index.bs
```diff
@@ -1139,12 +1141,17 @@ To <dfn>canonicalize a sanitizer element with attributes</dfn> a {{SanitizerElem
...
+   1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/exists=]:
+     1. [=list/iterate|For each=] |attribute| in
```
```suggestion
     1. [=list/iterate|For each=] |attribute| of
```

index.bs
```diff
@@ -1139,12 +1141,17 @@ To <dfn>canonicalize a sanitizer element with attributes</dfn> a {{SanitizerElem
...
+        |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]:
+        1. [=list/Append=] the result of [=canonicalize a sanitizer attribute=] with |attribute| to |result|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
+  1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] [=map/exists=]:
+     1. [=list/iterate|For each=] |attribute| in
```
```suggestion
     1. [=list/iterate|For each=] |attribute| of
```

index.bs
evilpie: So instead `« [] »`, you want `« »`? That doesn't seem right to me.

evilpie: Okay, I should not have doubted you of course: https://html.spec.whatwg.org/#drag-and-drop-processing-model:~:text=Let%20urls%20be%20%C2%AB%20%C2%BB%2E

The HTML spec also sometimes seems to use "Let elements be an empty [list](https://infra.spec.whatwg.org/#list).", which might be easier to understand? 

https://html.spec.whatwg.org/#drag-and-drop-processing-model:~:text=Let%20elements%20be%20an%20empty%20list%2E

annevk: I prefer using the syntax consistently so people will get used to it and don't have to wonder about which it is. See also https://infra.spec.whatwg.org/#lists

index.bs
```diff
@@ -1138,12 +1140,15 @@ To <dfn>canonicalize a sanitizer element with attributes</dfn> a {{SanitizerElem
...
 1. If |element| is a [=dictionary=]:
-   1. [=list/iterate|For each=] |attribute| in
-      |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]:
-      1. [=list/Append=] the result of [=canonicalize a sanitizer attribute=] with |attribute| to |result|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].
-   1. [=list/iterate|For each=] |attribute| in
-      |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]:
-      1. [=list/Append=] the result of [=canonicalize a sanitizer attribute=] with |attribute| to |result|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"].
+   1. If neither |element|["{{SanitizerConfig/attributes}}"] nor
+      |element|["{{SanitizerConfig/removeAttributes}}"] [=map/exist=], then [=map/set=]
+      |result|["{{SanitizerConfig/removeAttributes}}"] to &laquo; &raquo;.
```
annevk: These are all linking the wrong dictionary I think. They should all be `SanitizerElementNamespaceWithAttributes` instead of `SanitizerConfig`, no?

index.bs
```diff
@@ -1138,12 +1140,15 @@ To <dfn>canonicalize a sanitizer element with attributes</dfn> a {{SanitizerElem
...
+   1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/exists=]:
```
annevk: This can be "Otherwise, if" then. Although maybe that is a little weird as we can't use it for the last one. Or maybe we can because we already ensured they are mutually exclusive at this point?

evilpie: Local attributes and removeAttributes aren't mutually exclusive.

evilpie: We could _continue_?

annevk: Oh yeah, that'd be nice. https://infra.spec.whatwg.org/#iteration-continue

evilpie:
Okay. I think it's even more correct now, we first have to create the list, it doesn't exist yet.

annevk (approved):

evilpie:
Wait, continue makes no sense. This isn't a loop..

PR #335, opened 2025-10-06
Sort the result of get()

### LINKED ISSUE #330: Sort the result of get()
Currently the Sanitizer specification uses Ordered Sets/Maps for all its lists like `elements`, `attributes` etc. This means the insertion order is observable by calling the `get()` method. Because we already normalize the lists anyway, we don't really want anyone the rely on the order. For that reason we will instead sort the result of the `get()` method.

The sort order would be:
1. Namespace
  a. `null` namespace always comes first
  b. strings (sorted by [code-unit-less-than](https://infra.spec.whatwg.org/#code-unit-less-than))
2. Name  (sorted by [code-unit-less-than](https://infra.spec.whatwg.org/#code-unit-less-than)).

This is also motivated by being able to use unsorted sets in implementations and the assumption that using the get() method will be much less common than using modifiers or sanitization itself.

evilpie:
Fixes #330

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/335.html" title="Last updated on Oct 13, 2025, 11:10 AM UTC (91f73dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/335/be5bbb4...evilpie:91f73dc.html" title="Last updated on Oct 13, 2025, 11:10 AM UTC (91f73dc)">Diff</a>

### DIFF
--- a/builtins/safe-baseline-configuration-materialized.json
+++ b/builtins/safe-baseline-configuration-materialized.json
@@ -2,7 +2,7 @@
   "removeElements": [
     {
       "namespace": "http://www.w3.org/1999/xhtml",
-      "name": "script"
+      "name": "embed"
     },
     {
       "namespace": "http://www.w3.org/1999/xhtml",
@@ -18,7 +18,7 @@
     },
     {
       "namespace": "http://www.w3.org/1999/xhtml",
-      "name": "embed"
+      "name": "script"
     },
     {
       "namespace": "http://www.w3.org/2000/svg",
--- a/builtins/safe-baseline-configuration.json
+++ b/builtins/safe-baseline-configuration.json
@@ -2,7 +2,7 @@
   "removeElements": [
     {
       "namespace": "http://www.w3.org/1999/xhtml",
-      "name": "script"
+      "name": "embed"
     },
     {
       "namespace": "http://www.w3.org/1999/xhtml",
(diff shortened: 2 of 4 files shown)


index.bs
```diff
@@ -1214,6 +1241,18 @@ To <dfn for="SanitizerConfig">add</dfn> a |name| to a |list|, where |name| is
 
 </div>
 
+<div algorithm>
+An item |itemA| is <dfn for="SanitizerConfig">less than item</dfn> |itemB| if:
+
+1. If |itemA|["namespace"] is `null`:
```
```suggestion
1. If |itemA|["namespace"] is null:
```
(applies throughout)

index.bs
```diff
@@ -1214,6 +1241,18 @@ To <dfn for="SanitizerConfig">add</dfn> a |name| to a |list|, where |name| is
...
+    1. Return true, if |itemB|["namespace"] is not `null`.
```
annevk: You want to start with the if statement (applies throughout).

index.bs
```diff
@@ -304,7 +304,33 @@ method steps are:
 </div>
 
 <div algorithm>
-The <dfn for="Sanitizer" method export>get()</dfn> method steps are to return the value of [=this=]'s [=Sanitizer/configuration=].
+The <dfn for="Sanitizer" method export>get()</dfn> method steps are:
+
+1. Let |config| be a copy of [=this=]'s [=Sanitizer/configuration=].
```
annevk: Since we don't ever expose this's configuration directly I don't think this is needed indeed. We maybe also want to explain somewhere that this ordering allows for implementations to use unordered sets internally if they wanted to.

evilpie: I added a note.

PR #351, opened 2025-10-31
Improve and update the existing Security Considerations

### LINKED ISSUE #287: Document threat models
We should have https://docs.google.com/document/d/1MNxsRbo8FIgP0WOxb_edreCY2em_KN8ALBu0FiWw0KY/edit in a more prominent place. And maybe we also want to stipulate what is out of scope, such as one-click attacks.

We also want to distinguish here between XSS safety and the more encompassing model for "default".

mozfreddyb:
- **Improve and update the existing Security Considerations - fixes #287 **

***
<a href="https://pr-preview.s3.amazonaws.com/mozfreddyb/sanitizer-api/pull/351.html" title="Last updated on Nov 5, 2025, 7:58 AM UTC (d65965d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/351/84178c2...mozfreddyb:d65965d.html" title="Last updated on Nov 5, 2025, 7:58 AM UTC (d65965d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1463,8 +1463,8 @@ according to a configuration. The Sanitizer API ships a strict baseline,
 such that scripting can never be allowed (cf. [[#never-allowed]]).
 In addition to this, a default safe list includes further restrictions that may help
 prevent a wide range of undesirable effects. These include, elements that can override
-site-specific settings (e.g., `<meta>`), embed third-party content or change the
-layout of the page (e.g., `<img>`, `<style>`, `aria-` attributes).
+site-specific settings (e.g., `<meta>`), embed third-party content, or change the
+layout/semantics of the page (e.g., `<img>`, `<style>`, `aria-` attributes).
 In addition, we disallow element and attribute configurations that lead to outgoing HTTP
 requests, and form controls. A detailed list is given in [[#default-disallowed-elements]].
 


index.bs
```diff
@@ -1453,10 +1453,15 @@ URLs, is as follows:
...
-and doing so would be a bug in the threat model.
+by traversing supplied HTML content and removing elements and attributes
+according to a configuration. The Sanitizer API ships a strict baseline,
+such that scripting can never be allowed (cf. [[#never-allowed]]).
+In addition to this, a default safe list includes further restrictions that may help
+prevent a wide range of undesirable effects. These include, elements that can override
+site-specific settings (e.g., `<meta>`), embed third-party content or change the
+layout of the page (e.g., `<img>`, `<style>`).
+In addition, we disallow element and attribute configurations that lead to outgoing HTTP
+requests, and form controls. A detailed list is given in [[#default-disallowed-elements]].
```
annevk, line 1469: I think we should add a qualifier of "without a click" to the outgoing requests.

Also seems good to call out ARIA here.

index.bs
```diff
@@ -1453,10 +1453,15 @@ URLs, is as follows:
 # Security Considerations # {#security-considerations}
 
 The Sanitizer API is intended to prevent DOM-based Cross-Site Scripting
```
annevk: In the heading below we just use "XSS". We should probably be somewhat consistent in this document about that. And we might also need a reference for this term I suppose if we want to use it.

index.bs
```diff
@@ -1479,8 +1483,8 @@ application by naming elements through `id` or `name` attributes such that
 properties like `children` of an HTML element in the DOM are overshadowed by
 the malicious content.
 
-The Sanitizer API does not protect DOM clobbering attacks in its
-default state, but can be configured to remove `id` and `name` attributes.
+The Sanitizer API disallows the `id` and `name` attributes by default, but
```
annevk, line 1492: I think we should reference the default configuration here, or some such.

index.bs
```diff
@@ -1512,13 +1516,13 @@ into a different parent element. An example for carrying out such an attack
...
-The Sanitizer API offers only functions that turn a string into a node tree.
-The context is supplied implicitly by all sanitizer functions:
+The Sanitizer API offers functions that combine parsing, sanitization and insertion.
+As such, the parsing context is supplied implicitly by all sanitizer functions:
 `Element.setHTML()` uses the current element; `Document.parseHTML()` creates a
 new document. Therefore Sanitizer API is not directly affected by mutated XSS.
 
 If a developer were to retrieve a sanitized node tree as a string, e.g. via
-`.innerHTML`, and to then parse it again then mutated XSS may occur.
+`.innerHTML`, to only parse or insert it again, a mutated XSS may still occur.
```
```suggestion
`.innerHTML`, to only parse or insert it again, mutated XSS may still occur.
```

index.bs
```diff
@@ -1527,6 +1531,483 @@ longer be considered as sanitized.
...
+    </tr>
+  </table>
+
+
+## Event Handler Content Attributes
+
+All known event handler content attributes are disallowed, as they execute script.
+
+ <table>
+   <caption>List of all known and therefore disallowed event handler content attributes attributes</caption>
```
annevk: I don't think we want to have yet another list for these. Let's use a reference instead.

index.bs
```diff
@@ -1527,6 +1531,483 @@ longer be considered as sanitized.
...
+
+
+: `<audio>`, `<canvas>`, `<embed>`, `<iframe>`, `<img>`, `<picture>`, `<video>`
+:: These elements for embedding content are not allowed, because it leads to HTTP requests to arbitrary hosts.
+
+: `<math>` MathML math
+:: The reasoning for the inclusion of MathML is solely based on [[SafeMathML]].
+
+: `<svg>`
+:: This element is not allowed, because it leads to HTTP requests to arbitrary hosts.
```
annevk: I thought we did allow some Math and SVG by default?

index.bs
```diff
@@ -1527,6 +1531,483 @@ longer be considered as sanitized.
...
+
+<em>This section is not normative.</em>
+
+## Elements and attributes not included in the default config ##  {#default-disallowed-elements}
+
+While the Sanitizer aims to disallow Cross-Site Scripting attacks, the default
+settings are a bit stricter and do not include elements that may change page
+settings, layout or fetch resources from other origins. The following elements
+can be allowed with a configuration, but the individual reasoning for disallowing
+them by default is given below.
```
annevk, line 1939: This list seems very incomplete. I think it either needs to cover all elements and attributes we made a decision on or we should not have it.

evilpie: +1. This was my first reaction after looking at this list as well.

index.bs
```diff
@@ -1527,6 +1531,483 @@ longer be considered as sanitized.
...
+
+See also: [[#security-considerations]].
+
+: `<meta>`
+:: This element is not allowed, because it can set or modify global document state.
+
+: `<link>`
+:: This element is not allowed, because it can set or modify global document state.
+
+: `<a target>`
```
evilpie: Mixing elements and attributes into one list makes it hard to scan.

index.bs
mozfreddyb: Is MDN an acceptable reference for web specifications?

index.bs
mozfreddyb: - MathML is some allowed as can only be seen when following the reference. I'll make it explicit
- You're right and I think I got that wrong. Will correct this in a follow-up.

index.bs
mozfreddyb: Good point. Some of the attributes were per-element and some were global. Let me try to strike a balance here in a follow-up

index.bs
annevk: Maybe, because it's only for non-normative text. But it gets kinda circular and doesn't seem ideal.

mozfreddyb: do you think we need to define it ourselves instead?

index.bs
mozfreddyb: The PR is getting a bit too unwieldy for my personal preference. I'll do the SVG/MATHML/HTML tables in a follow-up

index.bs
```diff
@@ -1452,11 +1457,17 @@ URLs, is as follows:
...
-the construction of a Sanitizer object that leaves script-capable markup in
-and doing so would be a bug in the threat model.
+The Sanitizer API is intended to prevent DOM-based Cross-Site Scripting [[XSS]]
+by traversing supplied HTML content and removing elements and attributes
+according to a configuration. The Sanitizer API ships a strict baseline,
+such that scripting can never be allowed (cf. [[#never-allowed]]).
+In addition to this, a default safe list includes further restrictions that may help
+prevent a wide range of undesirable effects. These include, elements that can override
+site-specific settings (e.g., `<meta>`), embed third-party content or change the
+layout of the page (e.g., `<img>`, `<style>`, `aria-` attributes).
```
annevk: ARIA doesn't embed or change the layout of the page.

annevk (commented):
I'm not sure how I feel about the "Index" section. It largely duplicates the existing lists, with some words added. That would probably be better inline?

index.bs
```diff
@@ -1527,6 +1537,479 @@ longer be considered as sanitized.
...
+    <tr>
+     <td><code>iframe</code></td>
+     <td>http://www.w3.org/1999/xhtml</td>
+     <td><a href="#html-el-1"><sup>1</sup></a></td>
+    </tr>
+
+    <tr>
+     <td><code>object</code></td>
+     <td>http://www.w3.org/1999/xhtml</td>
+     <td>This element is disallowed because it can include other, untrusted documents and including scripts.</td>
```
evilpie, line 1978: "including scripts" doesn't seem grammatically correct to me.

index.bs
```diff
@@ -1527,6 +1537,479 @@ longer be considered as sanitized.
...
+<li id="svg-attr-5">This attribute is disallowed because we do not allow the `<mask>` element.
+<li id="svg-attr-6">This attribute is disallowed because we do not allow the the element to impact the page layout.
+</ol>
+
+
+## Elements and attributes not allowed in the default config ##  {#default-disallowed-elements}
+
+While the Sanitizer aims to disallow Cross-Site Scripting attacks [[XSS]], the default
+settings are a bit stricter and do not include elements that may change page
+settings, layout or fetch resources from other origins. The following elements
```
evilpie, line 1937: We don't/can't really differentiate between fetching resources from same vs. other origins, so this seems too specific.

index.bs
```diff
@@ -1527,6 +1537,479 @@ longer be considered as sanitized.
...
+<li id="svg-attr-4">This attribute is disallowed because we do not allow the `<image>` element.
+<li id="svg-attr-5">This attribute is disallowed because we do not allow the `<mask>` element.
+<li id="svg-attr-6">This attribute is disallowed because we do not allow the the element to impact the page layout.
+</ol>
+
+
+## Elements and attributes not allowed in the default config ##  {#default-disallowed-elements}
+
+While the Sanitizer aims to disallow Cross-Site Scripting attacks [[XSS]], the default
+settings are a bit stricter and do not include elements that may change page
```
evilpie, line 1936: "a bit" feels a bit off here. Maybe "the default settings are purposefully stricter"

index.bs
```diff
@@ -1452,11 +1457,17 @@ URLs, is as follows:
 
 # Security Considerations # {#security-considerations}
 
-The Sanitizer API is intended to prevent DOM-based Cross-Site Scripting
-by traversing a supplied HTML content and removing elements and attributes
-according to a configuration. The specified API must not support
-the construction of a Sanitizer object that leaves script-capable markup in
-and doing so would be a bug in the threat model.
+The Sanitizer API is intended to prevent DOM-based Cross-Site Scripting [[XSS]]
+by traversing supplied HTML content and removing elements and attributes
```
otherdaniel, line 1461: Maybe clarify that sanitzation happens before inserting. E.g.:

[...] by traversing supplied HTML content, after parsing but before insertion into any document, and removing [...]

I think this is obvious to us, but I think a lot of people haven't entirely thought through the "when" aspect of sanitiaztion.

index.bs
```diff
@@ -1512,13 +1522,13 @@ into a different parent element. An example for carrying out such an attack
 is by relying on the change of parsing behavior for foreign content or
 mis-nested tags.
 
-The Sanitizer API offers only functions that turn a string into a node tree.
-The context is supplied implicitly by all sanitizer functions:
+The Sanitizer API offers functions that combine parsing, sanitization and insertion.
```
otherdaniel, line 1525: This is very true. I wonder if it bears saying at the top of this section, in the paragraph "The Sanitizer API is intended to [...]". Maybe instead of the comment I've proposed there.

Issue #354, opened 2025-11-05
Include additional reasoning for all elements not allowed in the default config


mozfreddyb:
I'll do a follow-up for #351.

Right now, I have a very long definition list that states every element that is not allowed in the default config with a long sentence (e.g., "this element is not allowed because ...").

This becomes hard to read, build, review and maintain for various reasons:
- Long repeated text is hard to scan and skim
- The spec builds on an allow list. A separate disallow-list in prose text essentially increases the burden to maintain two split lists that are disjunct
- There's no good way to maintain that the two lists added together are indeed the list of all known elements
- The definition list does not do proper justice to the attributes that we allow only on _some_ elements.

I will try and build a table of _all_ HTML elements based and go through those instead.

evilpie:
I would be fine with saying that all non-conforming elements are automatically disallowed by default.

Issue #349, opened 2025-10-30
Consider allowing the SVG <a> element


longsonr:
The sanitizer API allows the HTML `<a>` element but disallows the functionally equivalent SVG `<a>` element. Should it be consistent and allow both?

mozfreddyb:
True, it's not useful on its own. Some code outside of the sanitizer could switch between views? But I guess that case could be made for a lot of things we aren't allowing by default. We should err on the side of excluding when major features that would build on this (e.g., `<use>`) are also disallowed. People can still enable it in a config or use the unsafe sanitizer variant.


Let's stick to `<a>`.

otherdaniel:
> Looking at https://github.com/WICG/sanitizer-api/pull/298/files I don't think this was considered as "svg a" is not under "purposely omitted". This means there might be other elements for which that is true as well.

Yeah. I went through the doc chapter by chapter, and I guess skipped over chapter 16. That contains `<svg:a>`, and also <svg:view>`. Neither are listed in our doc; neither as allowed, nor as purposely omitted.

- `<a>`:  https://svgwg.org/svg2-draft/linking.html#Links
- `<view>`: https://svgwg.org/svg2-draft/linking.html#ViewElement

I'm not sure about `<view>`, but it seems to do largely the same thing as a `viewbox` attribute on <svg>, which we allow as well.

PR #355, opened 2025-11-05
Give allow/disallow reason in a big long tables (see #354)

mozfreddyb:
This builds on the fixes in #351 and the suggestion I made in #354.

I'm not entirely sure a long table is the best idea, but I also don't like any of the other options :)

***
<a href="https://pr-preview.s3.amazonaws.com/mozfreddyb/sanitizer-api/pull/355.html" title="Last updated on Nov 5, 2025, 8:59 AM UTC (b51c5c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/355/84178c2...mozfreddyb:b51c5c8.html" title="Last updated on Nov 5, 2025, 8:59 AM UTC (b51c5c8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1542,6 +1542,394 @@ A more complete treatment of mXSS can be found in [[MXSS]].
 
 <em>This section is not normative.</em>
 
+## Elements and attributes in the MathML namespace
+
+The `<math>` element and various additional elements are allowed.
+The list and their reasoning for the inclusion can be found in [[SafeMathML]].
+
+## Elements and attributes in the SVG namespace
+
+Features are either allowed, allowable (e.g., disallowed by default but can be included in a custom configuration) or completely disallowed.
+
+<table>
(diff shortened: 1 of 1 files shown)


otherdaniel:
I'd really like to have that list. Particularly in light of #349, where it turns out that we omitted something because of an oversight, and having a list which records the reasons allowed us to check that it was indeed an oversight. I wonder whether we should/could generate it; both for easier spec maintenance as well as for more flexible presentation.

I like how this enumerates the set of reasons. These basically form an enum of allowability. If we keep a file with name, namespace, allowability, we could quite easily generate these list. We could also derive the built-ins from them, or at least check them against the builtins for error checking.

---

Currently, we group by namespace, then list element/attr name -> allow-status. I suspect it'd be more compact if we'd group by namespace; then group by allow-status; then have a box of element/attr names.

Issue #323, opened 2025-09-25
Assert some of the expect Sanitizer state in the algorithms


evilpie:
@annevk  partly suggested this in #320 and I know @smaug--- is also interested in this.

Some ideas:
- Assert that dataAttributes exists iff attributes exists
- Assert that either global elements or removeElements exists
- Assert that either global attributes or removeAttributes exists

evilpie:
Actually, I just checked and it  probably wouldn't hurt to add this more explicitly before here: https://wicg.github.io/sanitizer-api/#sanitizerconfig-valid:~:text=If%20dataAttributes%20is%20true,-%3A.

PR #356, opened 2025-11-05
Make it explicit that dataAttributes needs to exist iff attributes exists

evilpie:
Part of #323.

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/356.html" title="Last updated on Nov 12, 2025, 9:59 AM UTC (e9419b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/356/dddc8df...evilpie:e9419b9.html" title="Last updated on Nov 12, 2025, 9:59 AM UTC (e9419b9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -758,6 +758,7 @@ conditions hold:
                 {{SanitizerConfig/dataAttributes}} is true:
                 1. |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] does not
                     contain a [=custom data attribute=].
+    1. [=Assert=]: |config|[{{SanitizerConfig/dataAttributes}}] [=map/exists=].
     1. If {{SanitizerConfig/dataAttributes}} is true:
         1. |config|[{{SanitizerConfig/attributes}}] does not contain
             a [=custom data attribute=].


otherdaniel:
Generally okay with thi.

> Or maybe this should be an assert? The canonicalization should be adding `dataAttributes` for this case automatically.

Not sure. I thought the intent of `[=SanitizerConfig/valid=]` is that it can be called on any, invalid SanitizerConfig instance. IMHO that speaks against an assert. On the other hand, we already assert that the names have been canonicalized, so arguably there's precedent in that same method. :)

Related: I think the WPT test for dataAttributes doesn't reflect the current spec anymore. It expects [`new Sanitizer({}).get().dataAttributes`](https://github.com/web-platform-tests/wpt/blob/master/sanitizer-api/sanitizer-boolean-defaults.tentative.html#L58) to be true, but I think the current spec interprets `{}` as  having `removeAttributes`, and that doesn't permit `dataAttributes`.

otherdaniel (approved):

PR #362, opened 2025-11-24
Rewrite `valid` as algorithm.

### LINKED ISSUE #313: Consider rephrasing "valid" as an algorithm
Currently we have statements such as:

> The |config| has either an {{SanitizerConfig/elements}} or a {{SanitizerConfig/removeElements}} key, but not both.

I think it would be clearer if we rewrote this and similar conditions to:

> If |config|["{{SanitizerConfig/elements}}"] exists and |config|["{{SanitizerConfig/removeElements}}"] exists, then return false.

otherdaniel:
This rewrite [=SanitizerConfig/valid=] as algorithm, instead of a set of conditions.
There should be no difference in semantics.

Fix: #313

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/362.html" title="Last updated on Feb 2, 2026, 7:04 PM UTC (0dcc39b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/362/f1d275e...otherdaniel:0dcc39b.html" title="Last updated on Feb 2, 2026, 7:04 PM UTC (0dcc39b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -738,26 +738,29 @@ NOTE: It's expected that the configuration being passing in has previously been 
     {{SanitizerAttributeNamespace}} items in |config| are canonical, meaning they have been run
     through [=canonicalize a sanitizer element=] or [=canonicalize a sanitizer attribute=],
     as appropriate.
-1. If |config|["{{SanitizerConfig/elements}}"] [=map/exists=] and
-    [=SanitizerConfig/has duplicates=], then return false.
-1. If |config|["{{SanitizerConfig/removeElements}}"] [=map/exists=] and
-    [=SanitizerConfig/has duplicates=], then return false.
+1. If |config|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+    1. If |config|["{{SanitizerConfig/elements}}"]
+        [=SanitizerConfig/has duplicates=], then return false.
+1. Otherwise:
+    1. If |config|["{{SanitizerConfig/removeElements}}"]
+        [=SanitizerConfig/has duplicates=], then return false.
 1. If |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/exists=] and
     [=SanitizerConfig/has duplicates=], then return false.
-1. If |config|["{{SanitizerConfig/attributes}}"] [=map/exists=] and
-    [=SanitizerConfig/has duplicates=], then return false.
-1. If |config|["{{SanitizerConfig/removeAttributes}}"] [=map/exists=] and
-    [=SanitizerConfig/has duplicates=], then return false.
-1. If |config|["{{SanitizerConfig/elements}}"] [=map/exists=],
-    |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/exists=], and
-    the [=SanitizerConfig/intersection=] of |config|["{{SanitizerConfig/elements}}"] and
-    |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] is not [=list/empty=], then return
-    false.
-1. If |config|["{{SanitizerConfig/removeElements}}"] [=map/exists=],
-    |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/exists=], and
-    the [=SanitizerConfig/intersection=] of |config|["{{SanitizerConfig/removeElements}}"] and
-    |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] is not [=list/empty=], then return
-    false.
+1. If |config|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
+    1. If |config|["{{SanitizerConfig/attributes}}"]
+        [=SanitizerConfig/has duplicates=], then return false.
+1. Otherwise:
+    1. If |config|["{{SanitizerConfig/removeAttributes}}"]
+        [=SanitizerConfig/has duplicates=], then return false.
+1. If |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/exists=]:
+    1. If |config|["{{SanitizerConfig/elements}}"] [=map/exists=]:
+        1. If the [=SanitizerConfig/intersection=] of |config|["{{SanitizerConfig/elements}}"] and
+            |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] is not [=list/empty=],
+            then return false.
+    1. Otherwise:
+        1. If the [=SanitizerConfig/intersection=] of |config|["{{SanitizerConfig/removeElements}}"] and
+            |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] is not [=list/empty=],
+            then return false.
 1. If |config|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
     1. [=Assert=]: |config|["{{SanitizerConfig/dataAttributes}}"] [=map/exists=].
     1. If |config|["{{SanitizerConfig/elements}}"] [=map/exists=]:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -719,63 +719,78 @@ Putting these rules in words:
...
 
 <div algorithm>
-A {{SanitizerConfig}} |config| is <dfn for=SanitizerConfig>valid</dfn> if all of the following
-conditions hold:
+To determine whether a {{SanitizerConfig}} |config| is <dfn for=SanitizerConfig>valid</dfn>:
 
-1. The |config| has either an {{SanitizerConfig/elements}} or a {{SanitizerConfig/removeElements}}
-    [=map/key=], but not both.
-1. The |config| has either an {{SanitizerConfig/attributes}} or a {{SanitizerConfig/removeAttributes}}
-    [=map/key=], but not both.
```
annevk, line 728: Did this move? (It seems like it got removed by accident.)

index.bs
```diff
@@ -719,63 +719,78 @@ Putting these rules in words:
...
+                &laquo; &raquo; is not [=list/empty=], then return false.
+            1. If |element|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}]
+                [=with default=] &laquo; &raquo; is not a
+                [=set/subset=] of |config|[{{SanitizerConfig/attributes}}], then return false.
             1. If {{SanitizerConfig/dataAttributes}} [=map/exists=] and
                 {{SanitizerConfig/dataAttributes}} is true:
-                1. |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] does not
-                    contain a [=custom data attribute=].
+                1. If |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}]
+                    contains a [=custom data attribute=], then return false.
```
annevk: Maybe combine this in a single longer conditional?

index.bs
```diff
@@ -719,63 +719,78 @@ Putting these rules in words:
...
     1. [=Assert=]: |config|[{{SanitizerConfig/dataAttributes}}] [=map/exists=].
     1. If {{SanitizerConfig/dataAttributes}} is true:
-        1. |config|[{{SanitizerConfig/attributes}}] does not contain
-            a [=custom data attribute=].
+        1. If |config|[{{SanitizerConfig/attributes}}] contains a [=custom data attribute=], then
+            return false.
```
annevk: Ditto.

annevk (commented):
Would be good if @evilpie did the final review here.

index.bs
otherdaniel: Done. (I indeed seem to have dropped it.)

index.bs

index.bs

index.bs
```diff
@@ -719,63 +719,80 @@ Putting these rules in words:
...
+            1. If |element|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}]
+                [=with default=] &laquo; &raquo; is not a
+                [=set/subset=] of |config|[{{SanitizerConfig/attributes}}], then return false.
             1. If {{SanitizerConfig/dataAttributes}} [=map/exists=] and
-                {{SanitizerConfig/dataAttributes}} is true:
-                1. |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] does not
-                    contain a [=custom data attribute=].
+                {{SanitizerConfig/dataAttributes}} is true and
+                |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}]
+                contains a [=custom data attribute=], then return false.
```
annevk, line 783: I can't make a suggestion here, but now the condition has "and" twice. First "and" should be replaced with a comma and the second needs a comma before it.

otherdaniel: Done.

annevk (approved):
Let me approve in case this holds up editing of other bits as Tom can also review post-merge I suppose.

index.bs
```diff
@@ -719,63 +719,80 @@ Putting these rules in words:
...
-A {{SanitizerConfig}} |config| is <dfn for=SanitizerConfig>valid</dfn> if all of the following
-conditions hold:
+To determine whether a {{SanitizerConfig}} |config| is <dfn for=SanitizerConfig>valid</dfn>:
 
-1. The |config| has either an {{SanitizerConfig/elements}} or a {{SanitizerConfig/removeElements}}
-    [=map/key=], but not both.
-1. The |config| has either an {{SanitizerConfig/attributes}} or a {{SanitizerConfig/removeAttributes}}
-    [=map/key=], but not both.
+1. If |config|[{{SanitizerConfig/elements}}] [=map/exists=] and
+    |config|[{{SanitizerConfig/removeElements}}] [=map/exists=], then return false.
```
evilpie: This is actually a pre-existing issue, but we use `|config|["{{SanitizerConfig/removeElements}}"]` (with quotes) everywhere else.

index.bs
```diff
@@ -719,63 +719,80 @@ Putting these rules in words:
...
-                [=set/subset=] of |config|[{{SanitizerConfig/attributes}}].
-            1. If {{SanitizerConfig/dataAttributes}} [=map/exists=] and
-                {{SanitizerConfig/dataAttributes}} is true:
-                1. |element|[{{SanitizerElementNamespaceWithAttributes/attributes}}] does not
-                    contain a [=custom data attribute=].
+                &laquo; &raquo; is not [=list/empty=], then return false.
+            1. If |element|[{{SanitizerElementNamespaceWithAttributes/removeAttributes}}]
+                [=with default=] &laquo; &raquo; is not a
+                [=set/subset=] of |config|[{{SanitizerConfig/attributes}}], then return false.
+            1. If {{SanitizerConfig/dataAttributes}} [=map/exists=],
```
evilpie: Also pre-existing: We already assert that it exists below.

index.bs
```diff
@@ -719,63 +719,80 @@ Putting these rules in words:
...
-conditions hold:
+To determine whether a {{SanitizerConfig}} |config| is <dfn for=SanitizerConfig>valid</dfn>:
 
-1. The |config| has either an {{SanitizerConfig/elements}} or a {{SanitizerConfig/removeElements}}
-    [=map/key=], but not both.
-1. The |config| has either an {{SanitizerConfig/attributes}} or a {{SanitizerConfig/removeAttributes}}
-    [=map/key=], but not both.
+1. If |config|[{{SanitizerConfig/elements}}] [=map/exists=] and
+    |config|[{{SanitizerConfig/removeElements}}] [=map/exists=], then return false.
+1. If |config|[{{SanitizerConfig/attributes}}] [=map/exists=] and
```
evilpie: A config without either attributes or removeAttributes isn't valid.

index.bs
```diff
@@ -719,63 +719,80 @@ Putting these rules in words:
...
-1. None of |config|[{{SanitizerConfig/elements}}], |config|[{{SanitizerConfig/removeElements}}],
-    |config|[{{SanitizerConfig/replaceWithChildrenElements}}],
-    |config|[{{SanitizerConfig/attributes}}], or
-    |config|[{{SanitizerConfig/removeAttributes}}], if they [=map/exist=],
-    [=SanitizerConfig/has duplicates=].
-1. If both |config|[{{SanitizerConfig/elements}}] and
-    |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exist=], then
+1. If |config|[{{SanitizerConfig/elements}}] [=map/exists=] and [=SanitizerConfig/has duplicates=],
+    then return false.
+1. If |config|[{{SanitizerConfig/removeElements}}] [=map/exists=] and
```
evilpie: I generally prefer:

1. If config["elements"] exists:
 1.1. If  elements has duplicates, then return false.
2. Otherwise,
 2.1 If removeElements has duplicates, then return false.

(Same for attributes/removeAttributes)

index.bs
```diff
@@ -719,63 +719,80 @@ Putting these rules in words:
...
+    [=SanitizerConfig/has duplicates=], then return false.
+1. If |config|[{{SanitizerConfig/elements}}] [=map/exists=],
+    |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exists=], and
     the [=SanitizerConfig/intersection=] of |config|[{{SanitizerConfig/elements}}] and
-    |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is [=list/empty=].
-1. If both |config|[{{SanitizerConfig/removeElements}}] and
-    |config|[{{SanitizerConfig/replaceWithChildrenElements}}] [=map/exist=], then
+    |config|[{{SanitizerConfig/replaceWithChildrenElements}}] is not [=list/empty=], then return
+    false.
+1. If |config|[{{SanitizerConfig/removeElements}}] [=map/exists=],
```
evilpie: I would think _and_ instead of _,_ ?

evilpie (changes requested):

index.bs
evilpie: Same applies to elements/removeElements of course.

index.bs
otherdaniel: Done.(Here & elsewhere.)

index.bs

index.bs
otherdaniel: Earlier feedback was to write multi-part conditionals as `a, b, and c`: https://github.com/WICG/sanitizer-api/pull/362#discussion_r2559674000

One is more idiomatic English; the other is more appropriate for formal logic. I'm be happy either way.

index.bs
otherdaniel: Hmm. I think that bit was just wrong. It shouldn't have been an assert. I now check for dataAttributes existance at the beginning of that block.

index.bs
otherdaniel: Hmm. I'd prefer to keep these, unless you insist.

Basically, I read these like the optional chaining ("?.") operator in JS: If x exist, then do stuff with it. That strikes me as reasonably compact and readable.

index.bs
evilpie: I believe we should assert everything that is enforced by the canonicalization?

index.bs
evilpie: I believe at this point I changed all other instances of this to follow my suggested pattern (#353), so for consistency it would be good to change.

index.bs
evilpie: I don't know. @annevk?

index.bs
evilpie: If we decide to assert everything is ensured by the canonicalization, we could do:

1. Assert: Either config["elements"] exists is true or config["removeElements"] exists is true.
2. If both ... are true, then return false.

I kind of like that we don't need to check something (but still assert) that should be given? In the current Step 3. we already assert that elements/attributes are canonical.


I find "If the exclusive or of" a bit awkward.

index.bs
annevk: "If a, b, and c, then Z" is correct.

evilpie (changes requested):
There are still some comments from my previous review.

index.bs

index.bs

evilpie:
I just looked at this again and it doesn't seem like my feedback was addressed.

evilpie (approved):

PR #369, opened 2026-01-29
Correctly remove attributes from the per-element attributes list

### LINKED ISSUE #368: "remove an attribute" does not always remove said attribute; and neither does "remove unsafe"
* [remove an attribute](https://wicg.github.io/sanitizer-api/#sanitizer-remove-an-attribute)
  * with a global attribute allow list and element allow lists,
  * will remove the given attribute from per-element attribute remove-lists, but *not* from per-element attribute allow-lists.
  * step 3.4.1.1
* This behaviour is tested by [sanitizer-modifiers](https://github.com/web-platform-tests/wpt/blob/master/sanitizer-api/sanitizer-modifiers.tentative.html#L170-L179)
* [remove unsafe](https://wicg.github.io/sanitizer-api/#sanitizerconfig-remove-unsafe) is defined in terms of [remove an attribute](https://wicg.github.io/sanitizer-api/#sanitizer-remove-an-attribute)
  * steps 5.1 + 6.1
* Thus, [remove unsafe](https://wicg.github.io/sanitizer-api/#sanitizerconfig-remove-unsafe) will not remove unsafe attributes, when they're
    given in a per-element allow list, when a global attribute allow list is present.

```js
// Returns: <html><body><p onclick="xxx"></p></body></html>. Note the event handler.
Document.parseHTML('<p onclick="xxx">', {sanitizer: {elements: [{ name:"p", attributes:["onclick"] }, "html", "body"], attributes:[]}});
```

----  
Original report: https://g-issues.chromium.org/issues/477643913

evilpie:
Fixes #368.

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/369.html" title="Last updated on Jan 29, 2026, 2:44 PM UTC (8744f41)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/369/f1d275e...evilpie:8744f41.html" title="Last updated on Jan 29, 2026, 2:44 PM UTC (8744f41)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1056,6 +1056,7 @@ to fix up per-element allow- or remove-lists to maintain our validity criteria. 
                     |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].            
             1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
                [=map/with default=] &laquo; &raquo; [=list/contains=] |attribute|:
+                1. [=Assert=]: |modified| is true.
                 1. [=list/Remove=] |attribute| from
                     |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"].
     1. Return |modified|.


index.bs
```diff
@@ -1043,18 +1043,22 @@ to fix up per-element allow- or remove-lists to maintain our validity criteria.
...
-        1. [=list/iterate|For each=] |element| in |configuration|["{{SanitizerConfig/elements}}"]:
+        1. [=list/iterate|For each=] |element| of |configuration|["{{SanitizerConfig/elements}}"]:
+            1. If |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+               [=map/with default=] &laquo; &raquo; [=list/contains=] |attribute|:
+                1. Set |modified| to true.
+                1. [=list/Remove=] |attribute| from
+                    |element|["{{SanitizerElementNamespaceWithAttributes/attributes}}"].            
             1. If |element|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"]
                [=map/with default=] &laquo; &raquo; [=list/contains=] |attribute|:
                 1. [=list/Remove=] |attribute| from
```
otherdaniel, line 1060: Would this also need to set |modified| to true?

evilpie: Not necessarily, but we could of course add it.

_modified_ must already be true, because the attribute must have been in _configuration_["attributes"] if it is in _element_["removeAttributes"].

annevk: We should add an assert if it must be true but is unclear.

annevk:
I think you should just go ahead Tom. It's an obvious bug fix. And the document is still in WICG so it's not all that formal yet.

mozfreddyb:
thanks for the patch and the quick review y'all, this is outstanding 🙌

PR #375, opened 2026-03-04
Replace `replace all` with steps from `replaceWith`.

### LINKED ISSUE #372: Incorrect use of "replace all"
[This use of replace all](https://wicg.github.io/sanitizer-api/#config:~:text=Call%20replace%20all%20with%20child%E2%80%99s%20children%20within%20child%2E) is incorrect.

"replace all" replaces all children of an element. Here it's replacing all children of an element with the children of the element, resulting in no change.

Instead, it should be following steps similar to [`replaceWith`](https://dom.spec.whatwg.org/#dom-childnode-replacewith).

otherdaniel:
Replace `[=replace all=]` with `[=convert nodes into a node=]` and `[=replace=]`.

Fixes: #372

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/375.html" title="Last updated on Mar 23, 2026, 2:26 PM UTC (1c09b59)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/375/32bfe71...otherdaniel:1c09b59.html" title="Last updated on Mar 23, 2026, 2:26 PM UTC (1c09b59)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -32,6 +32,7 @@ text: HTML fragment parsing algorithm; type: dfn; url: https://html.spec.whatwg.
 text: global attribute; type: dfn; url: https://html.spec.whatwg.org/multipage/dom.html#global-attributes
 text: custom data attribute; type: dfn; url: https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute
 text: report a warning to the console; type: dfn; url: https://console.spec.whatwg.org/#report-a-warning-to-the-console
+text: convert nodes into a node; type: dfn; url: https://dom.spec.whatwg.org/#convert-nodes-into-a-node
 </pre>
 <pre class="biblio">
 {
@@ -913,7 +914,9 @@ beginning with |node|. It consistes of these steps:
       1. [=Assert=]: |node| does not [=implement=] {{Document}}.
       1. Call [=sanitize core=] on |child| with |configuration| and
           |handleJavascriptNavigationUrls|.
-      1. Call [=replace all=] with |child|'s [=tree/children=] within |child|.
+      1. Let |fragment| be the result of [=convert nodes into a node=]
+         given |child|'s [=tree/children=] and |node|'s [=node document=].
+      1. [=/Replace=] |child| with |fragment| within |child|'s [=tree/parent=].
       1. [=Continue=].
     1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=]:
       1. If |configuration|["{{SanitizerConfig/elements}}"] does not
--- a/index.bs
+++ b/index.bs
@@ -914,9 +914,16 @@ beginning with |node|. It consistes of these steps:
       1. [=Assert=]: |node| does not [=implement=] {{Document}}.
       1. Call [=sanitize core=] on |child| with |configuration| and
           |handleJavascriptNavigationUrls|.
-      1. Let |fragment| be the result of [=convert nodes into a node=]
-         given |child|'s [=tree/children=] and |node|'s [=node document=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -904,7 +905,9 @@ beginning with |node|. It consistes of these steps:
        [=SanitizerConfig/contains=] |elementName|:
       1. Call [=sanitize core=] on |child| with |configuration| and
```
annevk, line 915: Why is this needed?

index.bs
```diff
@@ -904,7 +905,9 @@ beginning with |node|. It consistes of these steps:
...
           |handleJavascriptNavigationUrls|.
-      1. Call [=replace all=] with |child|'s [=tree/children=] within |child|.
+      1. Let |fragment| be the result of [=convert nodes into a node=]
+         given |child|'s [=tree/children=] and |node|'s [=node document=].
+      1. [=/Replace=] |child| with |fragment| within |child|'s [=tree/parent=].
```
annevk: Instead of _child_'s parent should we just say _node_?

annevk: Also, we want to be somehow clear here that we don't expect this to throw. Maybe just with a note.

index.bs
```diff
@@ -904,7 +905,9 @@ beginning with |node|. It consistes of these steps:
        [=SanitizerConfig/contains=] |elementName|:
       1. Call [=sanitize core=] on |child| with |configuration| and
           |handleJavascriptNavigationUrls|.
-      1. Call [=replace all=] with |child|'s [=tree/children=] within |child|.
+      1. Let |fragment| be the result of [=convert nodes into a node=]
+         given |child|'s [=tree/children=] and |node|'s [=node document=].
```
annevk: I think it might be slightly better if we just build up the fragment ourselves.

1. Let _fragment_ be a new DocumentFragment node whose node document is _node_'s node document.
2. For each _innerChild_ of _child_'s children: append _innerChild_ to _fragment_.

Because "convert nodes into a node" does some things we don't really need here.

(Maybe even better would be updating the DOM standard with a new abstract operation just for that.)

index.bs
otherdaniel: My assumption was that the top-level `list/iterate` would have some sort of pass-by-value semantics, i.e. that the list at the time of the iteration is fixed. In that case this is needed, because otherwise the sanitizer would never see any of the children that are replacing their parents.

Spec text [says](https://infra.spec.whatwg.org/#list-iterate):
> To iterate over a [list](https://infra.spec.whatwg.org/#list), performing a set of steps on each [item](https://infra.spec.whatwg.org/#list-item) in order, [...]

That doesn't strike me as very clear either way.

I also found [this discussion](https://github.com/whatwg/infra/issues/396), which seems to resolve this in favour of supporting modifications while iterating. But the promised PR from that discussion seems to not have materialized.

Not sure what to do here... by my reading of the current spec texts, I can't find a guarantee that modify during iterate would work as expected.

index.bs

index.bs

index.bs
zcorpan: https://github.com/whatwg/infra/pull/451 exists but seems to have stalled.

otherdaniel:
@noamr fyi: This PR was landed after https://github.com/whatwg/html/pull/12395 was written.

noamr:
> @noamr fyi: This PR was landed after [whatwg/html#12395](https://github.com/whatwg/html/pull/12395) was written.

Assimilated into the upstreaming PR, thanks!

PR #377, opened 2026-03-04
Reject config with <html> in replaceWithChildrenElements as invalid

### LINKED ISSUE #365: Don't allow multiple root nodes to be created when using `replaceWithChildren: ["html"]`
The HTML/DOM spec doesn't allow you to create a document with multiple root nodes, e.g. calling `appendChild` on a document throws.

However during fuzzing we found a test case that shows it would be possible to do this using the Sanitizer API:
```
Document.parseHTML("<div>1</div><div>2</div>",  { sanitizer: { replaceWithChildrenElements: ["html"] } })
```

It also seems like doing `element.setHTML("<html>abc</html>")` drops the HTML tag already.

So the easiest solution would be to just not allow the HTML element inside `replaceWithChildrenElements` as part of the config during validation.

evilpie:
Fixes #365.

We can't allow the replacement of the root `<html>` element with multiple elements. So we reject any config that has `replaceWithChildrenElements: ["html"]`.

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/377.html" title="Last updated on Mar 9, 2026, 3:30 PM UTC (c92d8ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/377/84c7deb...evilpie:c92d8ce.html" title="Last updated on Mar 9, 2026, 3:30 PM UTC (c92d8ce)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -441,6 +441,9 @@ The <dfn for="Sanitizer" method export>replaceElementWithChildren(|element|)</df
 1. Let |configuration| be [=this=]'s [=Sanitizer/configuration=].
 1. [=Assert=]: |configuration| is [=SanitizerConfig/valid=].
 1. Set |element| to the result of [=canonicalize a sanitizer element=] with |element|.
+1. If |element|["{{SanitizerElementNamespace/name}}"] [=string/is|equals=] "`html`"
+    and |element|["{{SanitizerElementNamespace/namespace}}"] [=string/is|equals=] [=HTML namespace=]:
+    1. Return false.
 1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/contains=] |element|:
     1. Return false.
 1. [=SanitizerConfig/Remove=] |element| from |configuration|["{{SanitizerConfig/removeElements}}"].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -753,6 +753,9 @@ NOTE: It's expected that the configuration being passing in has previously been
     1. If |config|["{{SanitizerConfig/removeAttributes}}"]
         [=SanitizerConfig/has duplicates=], then return false.
 1. If |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/exists=]:
+    1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"]
```
otherdaniel: I think this also needs an extra step in the `replaceWithChildren` method.

An `[=Assert=]` in "sanitize core" that, when  `replaceWithChildrenElements` applies, checks whether parent is a Document node, might be nice, but is optional. That's the condition we're trying to protect.

evilpie: Oh of course, good catch.

index.bs
```diff
@@ -902,6 +908,7 @@ beginning with |node|. It consistes of these steps:
     1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/exists=]
        and if |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"]
        [=SanitizerConfig/contains=] |elementName|:
+      1. [=Assert=]: |node| does not [=implement=] {{Document}}.
```
evilpie, line 913: Does that make sense?

otherdaniel: Yes, I think that's the right condition.

otherdaniel (approved):
Looks good.

I think we need to also mention this at least once somewhere in the prose. E.g. in https://wicg.github.io/sanitizer-api/#invariants , maybe getting it's own bullet in the list following "Several conditions need to hold for a configuration to be valid:"

PR #385, opened 2026-03-23
Refactor "is attribute allowed"

### LINKED ISSUE #380: Attribute allowing conditions seem wrong
Step 1.5.9.3 of "sanitize core" reads:

> If configuration["[attributes](https://wicg.github.io/sanitizer-api/#dom-sanitizerconfig-attributes)"] does not [contain](https://wicg.github.io/sanitizer-api/#sanitizerconfig-contains) attrName and elementWithLocalAttributes["[attributes](https://wicg.github.io/sanitizer-api/#dom-sanitizerelementnamespacewithattributes-attributes)"] [with default](https://infra.spec.whatwg.org/#map-with-default) « » does not [contain](https://wicg.github.io/sanitizer-api/#sanitizerconfig-contains) attrName, and if "data-" is not a [code unit prefix](https://infra.spec.whatwg.org/#code-unit-prefix) of attribute’s [local name](https://dom.spec.whatwg.org/#concept-attribute-local-name) and [namespace](https://dom.spec.whatwg.org/#concept-attribute-namespace) is not null or configuration["[dataAttributes](https://wicg.github.io/sanitizer-api/#dom-sanitizerconfig-dataattributes)"] is not true:

In particular the following sub-condition:

> if "data-" is not a [code unit prefix](https://infra.spec.whatwg.org/#code-unit-prefix) of attribute’s [local name](https://dom.spec.whatwg.org/#concept-attribute-local-name) and [namespace](https://dom.spec.whatwg.org/#concept-attribute-namespace) is not null

Seems wrong to me. Would this not allow an attribute whose local name does not begin with "data-" but whose namespace is null, when data attributes are allowed? It seems to me like "and" should be "or".

otherdaniel:
Split the dreaded multi-line condition into several subclauses, assign each to an appropriately named boolean, and then test those bools.

That's more verbose than I usually go for, but it's surely more readable than the old version.

Fix: whatwg/sanitizer-api#380

***
<a href="https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/385.html" title="Last updated on May 22, 2026, 9:15 AM UTC (93d8663)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/385/8b4bdb0...otherdaniel:93d8663.html" title="Last updated on May 22, 2026, 9:15 AM UTC (93d8663)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -758,7 +758,7 @@ NOTE: It's expected that the configuration being passing in has previously been 
     1. If |config|["{{SanitizerConfig/removeAttributes}}"]
         [=SanitizerConfig/has duplicates=], then return false.
 1. If |config|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=map/exists=]:
-    1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"]
+    1. If |config|["{{SanitizerConfig/replaceWithChildrenElements}}"]
        [=SanitizerConfig/contains=] &laquo;[ "`name`" &rightarrow; "`html`",
         "`namespace`" &rightarrow; [=HTML namespace=] ]&raquo;, then return false.
     1. If |config|["{{SanitizerConfig/elements}}"] [=map/exists=]:
@@ -932,41 +932,10 @@ beginning with |node|. It consistes of these steps:
     1. If |child| is a [=shadow host=],
        then call [=sanitize core=] on |child|'s [=Element/shadow root=] with
        |configuration| and |handleJavascriptNavigationUrls|.
-    1. Let |elementWithLocalAttributes| be &laquo; [] &raquo;.
-    1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=] and
-        |configuration|["{{SanitizerConfig/elements}}"] [=SanitizerConfig/contains=]
-        |elementName|:
-       1. Set |elementWithLocalAttributes| to
-           |configuration|["{{SanitizerConfig/elements}}"][|elementName|].
     1. [=list/iterate|For each=] |attribute| in |child|'s [=Element/attribute list=]:
-      1. Let |attrName| be a {{SanitizerAttributeNamespace}} with |attribute|'s
-         [=Attr/local name=] and [=Attr/namespace=].
-
-      1. If |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] [=map/with default=] &laquo; &raquo;
-        [=SanitizerConfig/contains=] |attrName|:
-          1. [=/remove an attribute|Remove=] |attribute|.
-      1. Otherwise, if |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
-          1. Let the [=boolean=] |globallyAllowed| be whether
-              |configuration|["{{SanitizerConfig/attributes}}"]
-              [=SanitizerConfig/contains=] |attrName|.
-          1. Let the [=boolean=] |locallyAllowed| be whether
-              |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] [=map/with default=] &laquo; &raquo;
-              [=SanitizerConfig/contains=] |attrName|.
-          1. Let the [=boolean=] |isDataAttributeAllowed| be whether both,
-              "data-" is a [=code unit prefix=] of |attribute|'s
-              [=Attr/local name=] and [=Attr/namespace=] is `null`, and
-              |configuration|["{{SanitizerConfig/dataAttributes}}"] is true.
-          1. If neither |globallyAllowed| nor |locallyAllowed| nor
-              |isDataAttributeAllowed|,
-              [=/remove an attribute|remove=] |attribute|.
-      1. Otherwise:
-        1. If |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
-          [=map/exists=] and |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] does not [=SanitizerConfig/contain=] |attrName|:
-          1. [=/remove an attribute|Remove=] |attribute|.
-
-        1. Otherwise, if |configuration|["{{SanitizerConfig/removeAttributes}}"] [=SanitizerConfig/contains=] |attrName|:
-              1. [=/remove an attribute|Remove=] |attribute|.
-
+        1. If [=is attribute allowed=] for |attribute| given |configuration|
+            and |child| is `remove`,
+            then [=/remove an attribute|remove=] |attribute|.
       1. If |handleJavascriptNavigationUrls|:
          1. If &laquo;[|elementName|, |attrName|]&raquo; matches an entry in
             the [=built-in navigating URL attributes list=], and if |attribute|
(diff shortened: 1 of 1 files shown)


evilpie:
I have been thinking it might be cleaner to add a new algorithm like **is attribute allowed**. This might also be useful for #387, because we could use it like "is attribute allowed `is`".

otherdaniel:
> I have been thinking it might be cleaner to add a new algorithm like **is attribute allowed**. This might also be useful for #387, because we could use it like "is attribute allowed `is`".

Done. Please take another look.

index.bs
```diff
@@ -932,36 +932,10 @@ beginning with |node|. It consistes of these steps:
...
-      1. Otherwise:
-        1. If |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
-          [=map/exists=] and |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] does not [=SanitizerConfig/contain=] |attrName|:
-          1. [=/remove an attribute|Remove=] |attribute|.
-
-        1. Otherwise, if |configuration|["{{SanitizerConfig/removeAttributes}}"] [=SanitizerConfig/contains=] |attrName|:
-              1. [=/remove an attribute|Remove=] |attribute|.
-
+        1. If [=is attribute allowed=] for |attribute| given |configuration|
+            and |child| is `remove`,
```
annevk: I would suggest using a boolean or a pair of singletons: https://infra.spec.whatwg.org/#singleton

index.bs
```diff
@@ -980,6 +954,48 @@ beginning with |node|. It consistes of these steps:
...
 
+<div algorithm>
+To determine <dfn>is attribute allowed</dfn> for an |attribute|,
+given a {{SanitizerConfig}} |configuration| and an {{Element}} |current element|:
+
+1. Let |attrName| be a {{SanitizerAttributeNamespace}} with |attribute|'s
+    [=Attr/local name=] and [=Attr/namespace=].
+1. Let |elementName| be a {{SanitizerElementNamespace}} with |current element|'s
+    [=Element/local name=] and [=Element/namespace=].
+1. Let |elementWithLocalAttributes| be &laquo; [] &raquo;.
```
```suggestion
1. Let |elementWithLocalAttributes| be &laquo; &raquo;.
```

index.bs
```diff
@@ -980,6 +954,48 @@ beginning with |node|. It consistes of these steps:
...
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=] and
+    |configuration|["{{SanitizerConfig/elements}}"] [=SanitizerConfig/contains=]
+    |elementName|:
+   1. Set |elementWithLocalAttributes| to
+       |configuration|["{{SanitizerConfig/elements}}"][|elementName|].
```
annevk: It was a list but now it's an object?

index.bs
```diff
@@ -980,6 +954,48 @@ beginning with |node|. It consistes of these steps:
...
+1. If |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/removeAttributes}}"] [=map/with default=] &laquo; &raquo;
+    [=SanitizerConfig/contains=] |attrName|:
+    1. Return `remove`.
+1. Otherwise, if |configuration|["{{SanitizerConfig/attributes}}"] [=map/exists=]:
```
annevk: Do we need to say "Otherwise" here? We're using early returns after all.

evilpie:
Should we move the `handleJavascriptNavigationUrls` branch also into **is attribute allowed**?

index.bs
otherdaniel: Done. Using [=/blocked=] and [=/allowed=]

index.bs
otherdaniel: It was supposed to be an empty ordered map, per https://infra.spec.whatwg.org/#maps. Now I just use the phrase "empty ordered map", which is used elsewhere.

index.bs
otherdaniel: :)  It's now an empty ordered map or an IDL dictionary. I think I can use both with square brackets operator.

index.bs

otherdaniel:
> Should we move the `handleJavascriptNavigationUrls` branch also into **is attribute allowed**?

Done.

index.bs
```diff
@@ -932,54 +932,72 @@ beginning with |node|. It consistes of these steps:
...
+    [=Attr/local name=] and [=Attr/namespace=].
+1. Let |elementName| be a {{SanitizerElementNamespace}} with |current element|'s
+    [=Element/local name=] and [=Element/namespace=].
+1. Let |elementWithLocalAttributes| be an empty [=ordered map=].
+1. If |configuration|["{{SanitizerConfig/elements}}"] [=map/exists=] and
+    |configuration|["{{SanitizerConfig/elements}}"] [=SanitizerConfig/contains=]
+    |elementName|:
+   1. Set |elementWithLocalAttributes| to the |item| in
+       |configuration|["{{SanitizerConfig/elements}}"] where
+       |elementName|[{{SanitizerElementNamespace/name}}] [=string/is=]
```
evilpie: Shouldn't this (and below) be `|elementName|["{{SanitizerElementNamespace/name}}"]` (with quotes)?

index.bs
```diff
@@ -932,54 +932,72 @@ beginning with |node|. It consistes of these steps:
...
+1. Otherwise:
+    1. If |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+        [=map/exists=] and |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] does not [=SanitizerConfig/contain=] |attrName|:
+        1. Return [=/blocked=].
+1. If |handleJavascriptNavigationUrls|:
+   1. If &laquo;[|elementName|, |attrName|]&raquo; matches an entry in
+      the [=built-in navigating URL attributes list=], and if |attribute|
+      [=contains a javascript: URL=], then return [=/blocked=].
+   1. If |current element|'s [=Element/namespace=] [=string/is=] the
+      [=MathML Namespace=] and |attr|'s [=Attr/local name=] [=string/is=]
```
evilpie:
```suggestion
      [=MathML Namespace=] and |attribute|'s [=Attr/local name=] [=string/is=]
```

(and more)

evilpie (changes requested):

evilpie:
Sorry, I just realized that this won't really work for my idea of using this for implementing a manual `is` check, because in that case we don't have a real `Attr attr` attribute.

index.bs
```diff
@@ -932,54 +932,72 @@ beginning with |node|. It consistes of these steps:
...
+        "data-" is a [=code unit prefix=] of |attribute|'s
+        [=Attr/local name=] and [=Attr/namespace=] is `null`, and
+        |configuration|["{{SanitizerConfig/dataAttributes}}"] is true.
+    1. If neither |globallyAllowed| nor |locallyAllowed| nor
+        |isDataAttributeAllowed|, return [=/blocked=].
+1. Otherwise:
+    1. If |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+        [=map/exists=] and |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] does not [=SanitizerConfig/contain=] |attrName|:
+        1. Return [=/blocked=].
+1. If |handleJavascriptNavigationUrls|:
```
evilpie: Sorry for flip-flopping on this, but please remove this again ... For #396.

evilpie:
@otherdaniel Can I help getting this done?

index.bs

index.bs

index.bs
otherdaniel: Moved back into 'sanitize core'.

otherdaniel:
> @otherdaniel Can I help getting this done?

Please give it another look. Apologies for being a bit late with this.

index.bs
```diff
@@ -1032,54 +1032,75 @@ beginning with |node|. It consistes of these steps:
...
-            [=MathML Namespace=] and |attr|'s [=Attr/local name=] [=string/is=]
-            "`href`" and  |attr|'s [=Attr/namespace=] is `null` or the
-            [=XLink namespace=] and |attr| [=contains a javascript: URL=],
-            then [=/remove an attribute|remove=] |attribute|.
-         1. If the [=built-in animating URL attributes list=]
-            [=SanitizerConfig/contains=]
-            «[|elementName|, |attrName|]» and |attr|'s
-            [=get an attribute value|value=] [=string/is=] "`href`" or
-            "`xlink:href`", then [=/remove an attribute|remove=] |attribute|.
+        1. Let |attrName| be a {{SanitizerAttributeNamespace}} with |attribute|'s
```
evilpie: Let's move this before **is attribute allowed**, and then call it with `|attrName|`.

otherdaniel: Done. (Also calling it with |elementName| for consistency.)

otherdaniel: The error in the preview script seems unrelated to this change. If I've traced it correctly, `bikeshed` seems to expect boilerplate files in `bikeshed/spec-data/boilerplate`, but they're actually installed in `bikeshed/spec-data/readonly/boilerplate`. No idea why...

index.bs
```diff
@@ -1032,54 +1032,73 @@ beginning with |node|. It consistes of these steps:
...
+        "data-" is a [=code unit prefix=] of
+        |attrName|[{{SanitizerAttributeNamespace/name}}] and
+        |attrName|[{{SanitizerAttributeNamespace/namespace}}] is `null`, and
+        |configuration|["{{SanitizerConfig/dataAttributes}}"] is true.
+    1. If neither |globallyAllowed| nor |locallyAllowed| nor
+        |isDataAttributeAllowed|, return [=/blocked=].
+1. Otherwise:
+    1. If |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"]
+        [=map/exists=] and |elementWithLocalAttributes|["{{SanitizerElementNamespaceWithAttributes/attributes}}"] does not [=SanitizerConfig/contain=] |attrName|:
+        1. Return [=/blocked=].
```
evilpie, line 1096: You seem to have dropped the check for SanitizerConfig/removeAttributes?

evilpie (changes requested):

index.bs
otherdaniel: You're right! Should now be fixed.

index.bs
```diff
@@ -1032,54 +1032,76 @@ beginning with |node|. It consistes of these steps:
...
+    1. Otherwise, if |configuration|["{{SanitizerConfig/removeAttributes}}"]
```
evilpie:
```suggestion
    1. If |configuration|["{{SanitizerConfig/removeAttributes}}"]
```

Early return now.

evilpie (approved):

evilpie:
@noamr fyi for upstreaming

noamr:
Since this is editorial perhaps upstreaming this change can be a follow up? It's getting difficult to keep track of the giant upstreaming PR

Issue #383, opened 2026-03-18
Don't allow replaceWithChildrenElements for integration points


evilpie:
Basically the same idea as #382.  Parsing behavior changes at [HTML integration point](https://html.spec.whatwg.org/multipage/parsing.html#html-integration-point) and [MathML text integration point](https://html.spec.whatwg.org/multipage/parsing.html#mathml-text-integration-point).

mozfreddyb:
Some findings:
In contrast to integration points, there are way to "escape" integration in the [The rules for parsing tokens in foreign content](https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inforeign)

mozfreddyb:
Example test case`<svg><font color="red"><clippath>`, whith the attribute "color" being removed by the sanitizer.

During parsing, the "color" being present makes the following elements *escape* SVG. If you sanitizer during parsing, what happens?

If you sanitize on insertion, we will parse as we already do (not sure)?

PR #394, opened 2026-04-15
Don't allow <svg> or <math> in replaceWithChildrenElements

### LINKED ISSUE #382: Don't allow replaceWithChildrenElements for <svg> and <math>
The `<svg>` and `<math>` elements switch the HTML parser to "foreign content" parsing mode and they introduce a new namespace.

When replacing either element with its children, it's trivial to generate a node tree that when serialized will be re-parsed differently.

Example:
```javascript
div.setHTML(`<svg><iframe src="data:text/html,hi"></iframe></svg>`, {
    sanitizer: {
        replaceWithChildrenElements: [{
            name: "svg",
            namespace: "http://www.w3.org/2000/svg"
        }]
    }
})
div.innerHTML = div.innerHTML; // `<iframe src="data:text/html,hi"></iframe>`
```

An `<iframe>` element in the SVG namespace has no special meaning, so we don't remove it. But after reparsing as HTML we will generate
a normal iframe element.

While we recognize that it might be hard to prevent every kind of re-parsing attack, "destroying" these namespace boundaries is just unnecessary.

Furthermore this nicely extends our existing approach of disallowing `<html>` in `replaceWithChildrenElements` for other reasons (#365).

@cure53 Mario, this issue might be interesting to you as well.

evilpie:
Fixes #382

We decided that we shouldn't allow `replaceWithChildrenElement` for `<svg>` or `<math>` because those introduce a new namespace and thus replacing them with children immediately leads to re-parsing issues.

I added  a new list **built-in non-replaceable elements list** which can easily be extended for #383.

***
<a href="https://pr-preview.s3.amazonaws.com/evilpie/sanitizer-api/pull/394.html" title="Last updated on Apr 15, 2026, 2:55 PM UTC (dfc1525)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/sanitizer-api/394/1f85a4d...evilpie:dfc1525.html" title="Last updated on Apr 15, 2026, 2:55 PM UTC (dfc1525)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -441,8 +441,7 @@ The <dfn for="Sanitizer" method export>replaceElementWithChildren(|element|)</df
 1. Let |configuration| be [=this=]'s [=Sanitizer/configuration=].
 1. [=Assert=]: |configuration| is [=SanitizerConfig/valid=].
 1. Set |element| to the result of [=canonicalize a sanitizer element=] with |element|.
-1. If |element|["{{SanitizerElementNamespace/name}}"] [=string/is|equals=] "`html`"
-    and |element|["{{SanitizerElementNamespace/namespace}}"] [=string/is|equals=] [=HTML namespace=]:
+1. If the [=built-in non-replaceable elements list=] [=SanitizerConfig/contains=] |element|:
     1. Return false.
 1. If |configuration|["{{SanitizerConfig/replaceWithChildrenElements}}"] [=SanitizerConfig/contains=] |element|:
     1. Return false.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -624,8 +623,9 @@ Several conditions need to hold for a configuration to be valid:
 - Duplicate entries on the same element:
   - There are no duplicate entries between {{SanitizerElementNamespaceWithAttributes/attributes}}
     and {{SanitizerElementNamespaceWithAttributes/removeAttributes}} on the same element.
-- The HTML `<html>` element must not appear in {{SanitizerConfig/replaceWithChildrenElements}},
-  since replacing it with its children could lead to an invalid state with multiple root nodes.
+- No element from the [=built-in non-replaceable elements list=] appears in {{SanitizerConfig/replaceWithChildrenElements}},
+  since replacing these elements with their children could lead to re-parsing issues
+  or invalid DOM trees.
```
annevk: node trees*

I was also discussing this issue with @freddyb some more over lunch and we should probably point out that we're not fixing all re-parsing issues. E.g., we don't plan to defend against `<table><tbody><tr><td><div>` with a policy that replaces `<td>` with children. Perhaps we should give that as an explicit example here.

index.bs
```diff
@@ -1494,6 +1495,19 @@ URLs, is as follows:
   <br>
 ]&raquo;
 
+The <dfn>built-in non-replaceable elements list</dfn> contains elements that must not be
+replaced with their children, as doing so can lead to re-parsing issues or an invalid DOM tree.
```
annevk: DOM -> node

annevk (approved):

index.bs
evilpie: Great idea. This also gave me the idea we should maybe give an example like this:
```js
div.setHTML(`<svg><iframe src="https://example.com"></iframe></svg>`, { sanitizer: { replaceWithChildrenElements: [{ name: "svg", namespace: "http://www.w3.org/2000/svg"}]  } })
div.innerHTML = div.innerHTML;
```

mozfreddyb:
CCing @noamr and @otherdaniel for visibility. Let's merge this and add tests.

evilpie:
(I am going to add tests when landing this change in Firefox)

